PR #1, opened 2016-04-11
Add devo7e-256 target.

silpstream:
Hi Guys,

Could we officially add the ability to build targets for 7e transmitters that have been upgraded to RCT6 and RET6 processors?

For these targets, `MODULAR` has been removed. `OPTIMIZE_DFU` is still here as I can't quite figure out if we should remove it. Seems to work either way and I can't see an advantage to not having it.

Nothing else here has changed except for creating new targets and turning on more features in `target_defs.h`.

Targets added are: `TARGET=devo7e_256` and `TARGET=devo7e_512`
`target_def.h` changes are:
_add:_
```
  #define HAS_PERMANENT_TIMER 1
  #define HAS_DATALOG         1
  #define HAS_LAYOUT_EDITOR   1
```
_remove:_
`  #define NO_LANGUAGE_SUPPORT 1
`

Thanks!


goebish:
Approved.

PhracturedBlue:
I don't particularly like the implementation.
hacking devo7e stuff directly into src/Makefile and targets/devo/common/Makefile.inc is ugly.  main.c is modified for space-only changes which shouldn't be there and I'm not willing to support the devo7e_512 at this time since it brings no benefit over the 256.

Instead the implementation should be done like the devof4 (i.e. create a new target and reuse the 7e files where appropriate)

silpstream:
I'll change the code to reflect the coding conventions you guys prefer. However I can't really see any support problems with devo7e_512. It is there more for build convenience and choice. Supporting it along side the 256 branch doesn't require anything extra, and we can inform users that the current bootloader can be used on it but only supports upto 244kb. What extra work is there beyond allowing the directory to exist with a .ld file and pointers to the 256 files? Am I missing something?

PhracturedBlue:
Primarily because it adds no value.When it does add value, we can easily add it.  Until then I see no point.

PhracturedBlue:
Also, do not add the devo7e_256 to the Makefile.  The makefile controls the default builds (those that show up in the nightlies).  I will decide when and if to add it to the supported list.  I am wary that unsuspecting users will try to use it and end up with a bricked Tx which is why I am wary about it.  Just creating a new directory in 'target' is enough to enable the build

silpstream:
Ok. Can do. :)

silpstream:
I've reimplemented based on what you've mentioned above. Should be good to go unless there is something else you think I've missed. Thanks!

PhracturedBlue:
getting close.  You should be able to link channels.c in the makefile.  For capabilities.h, you should just include the devo7e version (can't link it but shouldn't duplicate its contents).  This makes it easier to modify the devo7e and have the 256 variant tag along for free

silpstream:
Done. I was initially thinking of adding in the pots mod that you made to those 2 files but we can cross that bridge later.

silpstream:
Thank you! :))

mwm:
For convenience, you do realize you can just do "gmake TARGET=custom" to build your own "custom" target, right?

It might be nice if we added a "custom" target line "target: make TARGET=custom" to the makefile, but that's not a big deal.

silpstream:
Yes I did. Thanks for mentioning it!

PR #3, opened 2016-04-11
Fix FlySky packet period

goebish:
Newer FlySky receivers don't work with a packet period of 1460us.  
Quoting hpnuts:  
"Turnigy 9X Flysky module v2 is 1533, New Flysky TX module for 9XR/9XR Pro is 1510 (the one using the internal antenna) , brand new V911 TX is 1490.
1460 does not work but increasing to 1480 works. I kept 1510 has a middle value and tested it on all my RXs: flysky RX 3&6&8 channels, V911 and F939."

### DIFF
--- a/src/protocol/flysky_a7105.c
+++ b/src/protocol/flysky_a7105.c
@@ -41,6 +41,8 @@
 #define BIND_COUNT 2500
 #endif
 
+#define PACKET_PERIOD 1510
+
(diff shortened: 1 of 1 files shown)


PhracturedBlue:
I will pull this and hope we can get some testing before the 5.0 release

goebish:
I checked with an older FlySky RX, it's still working fine.

PR #5, opened 2016-05-06
Restore inverse highlighting on 128x64 screens

hexfet:
Commit 8bdf8668 removed code that indicates selected box by inverting background.  This change restores the highlighting.

Other cosmetic changes on the Devo7e and 10 were also introduced with 8bdf but are not affected by this PR.
- Labels in page heading (top row) are centered instead of left-justified
- Top row underline missing on mixer list and mixer channel pages.
Maybe others...


PhracturedBlue:
I don't like this solution.  the code for color and black-and white should really be split.  The current conditionals are very confusing merging incompatible styles together.  I didn't realize that hadn't been done so I'll work on it, but I don't want to merge this pull request as it is.

hexfet:
No argument here. Wasn't sure of the full intent of the change, but tracked down this part to keep the FrskyX testing going.

PhracturedBlue:
this has been resolved now.  No need for this anymore

PR #6, opened 2016-05-08
Add protocol FrskyX, an implementation supporting Frsky X-series receivers

hexfet:
Thanks to midelic and planger for providing the FrskyX implementation, and to all the testers and collaborators in http://www.deviationtx.com/forum/protocol-development/5356-new-frskyx-protocol.

This PR adds protocol FryskyX (frskyx_cc2500.c) which implements the Frsky D16 over-the-air protocol, including s.port and hub telemetry.  It also updates protocol Frsky (frsky2way_cc2500.c) to correctly support hub telemetry.  The HAS_EXTENDED_TELEMETRY definition is added to reduce the number of telemetry values in builds with limited memory (7e).  It is applied to Frsky and DSM telemetry.

The frskyx_cc2500.c file contains the FrskyX protocol implemented by midelic and planger, ported to Deviation. This is the Frysky "D16" protocol compatible with the Frsky X-series receivers.  The frsky_d_telem.inc file contains the hub telemetry decoder and is #included by both frskyx_cc2500.c and frsky2way_cc2500.c.  This adds working hub telemetry support to the Deviation Frsky implementation.  The other files in the PR contain changes to support additional telemetry values.

The FrskyX protocol supports up to 12 channels. Fixed ID binding is supported to link the transmitter wtih specific receivers.

The following protocol options are supported.
Freq-fine: Frequency offset adjustment. Range -127 to 127. Adjusts for variances betweeen CC2500 modules. Usually offset of 0 or -41 is required, but full range should be tested if there are problems with binding or range.  Default 0.

AD2GAIN: The VOLT2 telemetry value (AIN input on X4R) is multiplied by this value divided by 100. Allows adjustment for external resistor divider network.  Default 100 (gain of 1). Range is 1 to 2000 (gain of 0.01 to 20.00).

Failsafe: The Frsky failsafe options are fully supported. If the channel failsafe (in mixer channel config) is set this value is sent to the receiver every 9 seconds.  The receiver will use these values in failsafe mode unless the protocol option is set to RX.

For channels with failsafe set to off, the default Failsafe protocol option "Hold" commands the receiver to hold the last received channel values when the receiver enters failsafe mode.  The "NoPulse" setting causes the receiver to not send any signal on PPM outputs (Testing on X8R showed SBUS values went to minimum, but SBUS behavior is not specified by the protocol).  The "RX" setting prevents Deviation from sending failsafe settings so the receiver will use whatever failsafe values have been stored in the receiver.

Telemetry
Supports receiver telemetry, s.port and hub telemetry sensors as well as GPS.  All Devo transmitters support the s.port Frsky telemetry on the main page and in alarms.  The 128x64 telemetry test page is updated with additional values.  The supported telemetry values and their location on the Telemetry monitor page on the Devo10 are:

Misc|Bat|Cells
|----|----|----|
|RSSI|VOLT1|CELL1|
|TEMP1|VOLT2|CELL2|
|TEMP2|VOLT3|CELL3|
|RPM|MIN CELL|CELL4|
|FUEL|ALL CELL|CELL5|
|ALTITUDE|VOLTA|CELL6|
|VARIO|CURRENT|DISCHARGE|

The ALTITUDE value is reported as Above Ground Level.  The ground level is set to the first altitude telemetry value received.
The following values are calculated from received telemetry.  MIN CELL is the lowest reported CELL value.  ALL CELL is the total of all reported CELL values.  The Discharge derived value is total battery discharge amount in milliAmp-hours.

Due to space limitations the Devo7e only supports non-s.port telemetry (RSSI, VOLT1 for FrskyX.  Frsky adds VOLT2.)

The Frsky s.port telemetry provides for connecting up to 16 sensors of the same type (e.g. battery voltage).  Deviation supports multiple sensors of the same type, but only one telemetry value is saved.  The value most recently received from all the sensors of the same type is reported.

The value array in the Telemetry struct is changed from u16 to s32 to support negative vario values and altitude measured in centimeters.

In the radio protocol the values in packet[21] are not fully understood.  The current implementation works well enough for control and telemetry.  There is likely a forward data channel in packet[22-27].

The protocol supports up to 16 channels with 12 bit resolution.  Packets are transmitted every 9ms.


PhracturedBlue:
It looks pretty good.  I don't really love the way 'frsky_d_telem.inc' is handled, but I don't have a better option right now.  may rename to ._c instead, as 'inc' has been used only for Makefiles so far.

Did you test with the 320x240 gui?  I see a bunch of change to the 128x64 gui, but nothing for the 320x240 one.

The change is quite large, so I'll want to spend a little more time staring at it before I merge it.

hexfet:
Couldn't think of a better way to share common code and be compatible with 7e modules.  Changing the extension makes sense.

The gui changes add the extended telemetry values to the 128x64 telemtest page, but those changes have not been made to the 320x240 telemtest page.  The 128x64 changes came mainly from mwm, and with discussions about wider telemetry changes I avoided adding them to the 320x240 telemetry test page.  I can add it but haven't looked yet.  Today the extended values can be placed in the 320x240 main page boxes and have been tested there by Alexandro.

PhracturedBlue:
I'd rather have consistent support across all transmitters at each point in time.  So please add the extra fields to the 320x240 gui and change the extension from inc to ._c and resubmit.

hexfet:
[How's this](https://drive.google.com/open?id=0B3AKcbg1PFrneERHUVdUSDdVMGs) for the 320x240 telemetry test page layout?  The only value missing compared to 128x64 is Cell6 (maybe that should replace Min Cell?).

PhracturedBlue:
It isn't pretty, but it'll do I think.  Thanks.

hexfet:
Yes, my aesthetic ranges from utilitarian to utilitarian.  Successful test reported in link in first post.

The hub telemetry is getting some more testing with Frysky (D8).  I've already got some changes but waiting for testing before pushing.

PR #9, opened 2016-05-17
devo7e-256k: Enable up to four 3-way switches and 2 pots.

silpstream:
Enables up to four 3-way switches and 2 pots. Configuration is via hardware.ini and options stack similar to devo6s extra buttons.

Adds the ability to select [3x3, 3x4] with [2x1, 2x2] as combinations for switches
Adds the ability to select [potx2, potx1].
Previous 2x2, 3x1, and 3x2 switch mod still works.

Emulator has been updated also for this mod. Configuration is similarly done via hardware.ini.


PhracturedBlue:
Note...this code will likely need to be updated to support the recent changes to the makefile.  Please verify that it builds properly on master.

silpstream:
Thanks for the heads up PB! Just finished some updates to support the recent changes. Tested both emu and tx on my end and it all checks out now.

PR #7, opened 2016-05-13
Minor changes to enable portability to atmega

pascallanger:

PR #14, opened 2016-05-28
AT9 Discrete Logic 4-in-1 module support and common-proto

victzh:
This pull request contains two sets of changes:

1. Pascal's pull request #7, modified so it builds for Devo 7E
2. Support for the board with 4-in-1 and SPI flash for AT9

The purpose of this pull request is to check the code builds and to integrate it later. Don't apply automatically.


PhracturedBlue:
It generally looks good.  I'm not sure about the at9-mod thing.  If I want to support this on all of the f* tx as well, it would mean a lot more targets and confusion for everyone.  I'm thinking it would make more sense to enable through the hardware.ini.  The big problem is---how do you know to use it instead?
We could choose to do like on the Devo12 and just have the media on the 2nd large drive, but that makes USB pretty annoying to work with.  Even better would be to detect the module directly, but we'd need to do this without breaking users without one.  I need to think on that some-more.  However, at an absolute minimum, please do not dulicate files between at9 and at9-mod.  See the devo7e-256 mod, and how files are reused either with a #include, or with changes to the Makefile.  AT9 support is not quite complete, and I dn't want to duplicate every change in 2 places.

victzh:
Detecting this module is relatively easy - check for direct connection of CYRF6936, then if failed check for it after using switching logic.

I'm not a fan of a separate target as well, the purpose of this pull request is for you to check that it's reasonable. I can merge it into main at9, but there is a slight problem - my build uses different media and it is hard to build depending on a definition in target_def.h. One media set is pbms for reduced color, another is full color Devo8 bmp set. Anyway, as a first step I'll reduce redundancy in files by referencing files in at9.

PhracturedBlue:
yeah, the media thing is a bit more complicated.  we have plenty of room on the AT9 so supporting both pbm and bmp in the firmware would be easy.  switching between them based on the hardware.ini will require that we stop hard-coding the extension and instead use a function to generate it.  I can work on adding support for that.

victzh:
Merged at9-mod into at9, now to compile for DL module support use
make TARGET=at9 HAS_4IN1_DL_SUPPORT=1

PhracturedBlue:
It looks pretty good.  I don't like the name HAS_4IN1_DL_SUPPORT though.  Can you change it to 'HAS_4IN1_FLASH' instead?  it isn't really adding support, it is changing the behavior.  I'd also like to share the spi_flash code between all stm32 CPUs but I won't hold up the merge for that.

victzh:
I planned to merge the spi_code eventually.

I have strange behavior with CYRF6936 - I can bind DSMX module but it never receives channel values. I can't even bind Devo models. All other RF chips work fine.

Also, the main screen stopped re-drawing correctly. I'm trying to figure is it a race between the flash and radio or not. It does not re-draw even if I set protocol to 'None' in which case there is no RF time interrupt activity.

PhracturedBlue:
have you looked at it with a logic analyzer?  It should be very clear if you are getting the proper signals to the board.  whether the board actually works is a different question of course.  As for the scren redraw, I'll try it myself tomorrow and see if I can see what is going on.  Even with an interrupt during redraw, I don't see why a screen redraw would be affected by SPI.  Have you tried compiling with your SPI code disabled to make sure that still works?

victzh:
Rebuilt it after make clean, reinstalled filesystem - it works. Did not check Devo protocol, but DSMX and S-FHSS work. I added an ability to turn on/off modules for debugging, can be removed or modified later.

PhracturedBlue:
Sorry, one last thing.  in protocol.c, what is this?
+#if 0 // defined(HAS_4IN1_FLASH) && HAS_4IN1_FLASH
 +    return 1;
 +#else

Is it just leftover for debugging?  can you either use a proper condition for the if, remove it, or document why it is like it is?

victzh:
It is temporary debugging feature - it allows to turn on/off individual modules from hardware.ini. I'm not sure how hardware.ini should function for this module.

victzh:
If you prefer, I can reinstate the condition which disables [modules] section of hardware.ini altogether.

PhracturedBlue:
for some reason, github didn't refresh the pull request after I pulled the proto merge part.  can you refresh  and push to see if you can get it to refresh.  it is hard to review when the diff isn't accurate

victzh:
It seems to be updated now.

PR #17, opened 2016-06-02
Add accelerometer calibration flag for JJRC H20 / H11D

goebish:
Push both stick bottom right to calibrate accelerometer on JJRC H20 / H11D


PhracturedBlue:
Can you also update the manual with this info as well?

goebish:
Oops, I just noticed there's no documentation at all for H8_3D protocol, I'll do it.

goebish:
Done.
https://github.com/DeviationTX/deviation-manual/pull/6

PR #20, opened 2016-06-06
Fix rf channel hopping for H8_3D protocol

goebish:
Fix rf channel hopping calculation for H8_3D protocol, as noticed by mjbudden:
http://www.deviationtx.com/forum/protocol-development/4724-jjrc-h20-hex-quad?start=160#50020


PhracturedBlue:
you probably want to push this change to the 5.0 branch too

goebish:
Ok, done.

PR #25, opened 2016-06-09
added protocol documentation for CX-10A/CX-10 blue

mossmann:


goebish:
Hi mossmann, what a coincidence ... I just got a H1 and I was trying to decode gfsk from a CX10 transmitter today (well, from Deviation...), I think I almost got it but not completely, could you share the scripts/grc/whatever you used to write this documentation please ?

goebish:
There's a small mistake in the documentation, CX-10A is using the same protocol than CX-10 "blue".
The "green" one is slightly different (no VID).

mossmann:
Dominic Spill and I did some stuff with Ubertooth One for our talk at
ToorCamp yesterday.  They gave away a CX-10A to every attendee.  Our
demo code is here:

https://github.com/dominicgs/sturdy-palm-tree/tree/cheerson

slides:

http://greatscottgadgets.com/slides/ossmann-spill-toorcamp-2016.pdf

goebish:
Thank you very much for that and everything you do :)

victzh:
Michael, it's a nice surprise to see you here. I used your HackRF One to reverse engineer XN297, and goebish filled many blanks in my initial effort. With this chip becoming more and more popular in cheap Chinese quadcopters it was very useful to have such a powerful tool to spy directly on air - no amount of SPI tracing would help you understand that they used their own packet structure - preamble, address, and content encoding - all are different from their prototype, nRF24L01.

Thanks again for this amazing tool!

PR #27, opened 2016-06-18
Cc2500revsal

vfbl:
This is for CC2500 RF modules with GDO0 connected to RX_EN of PA/LNA and GD02 to TX_EN i.e. inverted, onboard. I have 2 of those from ebay. And another deviationtx member hit the same issue found this useful too.   To correct the inversion, add a line containing "swap_control-cc2500 = 1" into hardware.ini.  Without the fix, connection dropped a few feet away even with TX set to maximum output.

### DIFF
--- a/src/config/hardware.c
+++ b/src/config/hardware.c
@@ -31,6 +31,7 @@ const char HAPTIC_ENABLE[] = "enable-haptic";
 static const char SECTION_MODULES[] = "modules";
 static const char MODULE_ENABLE_PIN[] = "enable";
 static const char MODULE_HAS_PA[] = "has_pa";
+static const char MODULE_SWAP_CONTROL[] = "swap_control";
 const char * const MODULE_NAME[TX_MODULE_LAST] = {
       [CYRF6936] = "CYRF6936",
       [A7105]    = "A7105",
--- a/src/config/hardware.c
+++ b/src/config/hardware.c
@@ -100,6 +100,15 @@ static int ini_handler(void* user, const char* section, const char* name, const 
                return 1;
             }
         }
+        if(MATCH_START(name, MODULE_SWAP_CONTROL)) {
+            int pin = get_module_index(name+sizeof(MODULE_SWAP_CONTROL));
+            if(pin >= 0) {
+               int v = value_int ? 1 : 0;
+               Transmitter.module_swap_control = (Transmitter.module_swap_control & ~(1 << pin)) | (v << pin);
+               return 1;
+            }
+        }
+ 
     }
     return 0;
 }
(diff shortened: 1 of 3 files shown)


PhracturedBlue:
please provide a description for why this is needed.  Also, I do not like the name 'swap_control' it is completely uninformative

vfbl:
This is for CC2500 RF modules with GDO0 connected to RX_EN of PA/LNA and GD02 to TX_EN i.e. inverted onboard.  I have 2 of those from ebay.  And another deviationtx member hit the same issue, who found this useful.   

I cannot think of a short and informative name.  What would you suggest.  I guess if we include this we need some form of release notes to info users. Another way is to add an example line in hardware.ini with comments.

goebish:
Maybe you could name it swap_palna-cc2500, also add
; swap_palna-cc2500 = 0
in src\fs\common\hardware.ini

That would also require a note in the module installation guide, but I think the wiki hasn't been migrated to github yet.
https://bitbucket.org/PhracturedBlue/deviation/wiki/Enabling_modules

PhracturedBlue:
I'm not going to accept this patch.  The problem you are solving needs a more general solution.  for instance, we need to be able to distinguish between the AWA24S and UYChina and DEVO Cyrf modules.
A more generic solution is to add a 'config' option to the modules section, so you can set:
config-cc2500 = 0x01
(which we would then define the options for.  It is actually less readable than what you have, but with proper documentation, it will be fine, and a lot more flexible

vfbl:
AWA24S and UYChina Cyrf modules, what are those?  Where is the modules section?  Should I wait until the config option is in there?

I actually have modd'ed both my modules by cutting rails and rewiring hardwarewise.  So no hurry for me.  Only the member who is using the firmware to swap the pin may not be able to use new nightly builds until we have something in the stream.  

Regards

goebish:
AWA24S & BUYChina are replacement cyrf6936 for stock module or MULTIMOD.
Actually I find your solution is acceptable (with proper name), but it is for PhracturedBlue to decide and have the last word ;)

PhracturedBlue:
I just checked in some code that should work for you.
you can add defines in tx.h for module capabilities.  I added 'CC2500_REVERSE_GD02' for you.
you can now use that in cc2500.c to detect this alternate wiring (I didn't add this part for you)

You would then enable it in hardware.ini via:
config-cc2500 = 0x01

(maybe I'll allow using text instead in the future, but for now we'll keep it simple)

vfbl:
Thanks.  I can see you have added config.   

How do I sync up my branch to your changes on the web?  Should I disgard the branch and start all over?

vfbl:
Let me start over with a new branch, instead of backing out the changes.

PR #30, opened 2016-06-20
Add config-cc2500 0x01 for CC2500 modules with GDO0 connected to RX_E…

vfbl:
…N and GDO2 connected to TX_EN of PA/LNA.

Add config-cc2500 = 0x01 to modules section of hardware.ini to enable reversal.

### DIFF
--- a/src/protocol/spi/cc2500.c
+++ b/src/protocol/spi/cc2500.c
@@ -96,15 +96,23 @@ void CC2500_WriteData(u8 *dpbuffer, u8 len)
 
 void CC2500_SetTxRxMode(enum TXRX_State mode)
 {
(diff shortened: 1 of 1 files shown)


vfbl:
Ar, just forgot to ask.  

Would you want to use the config byte as bit map?  Then we check if the bit is set instead?

PhracturedBlue:
I chose to use hex so we can use it as a bitmask if needed.  But with only one value it doesn't make any difference at this point.  My only request is that you use the CC2500_REVERSE_GD02 enum in the condition instead of '== 1' so that we have some idea of what it is supposed to do in 6 months time.

vfbl:
Sorry I forgot about CC2500_REVERSE_GD02.   Let me change it now.  Hope I can.  :)

PR #32, opened 2016-08-18
Adding a user option to disable Auto-Acknowledge on the YD717 Protocol's

NathanBarguss:
After swapping out the nrf24l01 in my 7e for the 4-in-1 module, I could no longer fly the Syma X7 without a huge amount of lag.

Some investigation later, and it turns out that Auto-Ack was the culprit, with the 4-in-1 not receiving the ACKs and wasting lots of time on re-transmits. 

A fast fix was to allow the user to selectively disable AA.

More details:
http://www.deviationtx.com/forum/6-general-discussions/6230-my-devo-7e-4-in1-2x3-build-and-syma-x4-issues

### DIFF
--- a/src/protocol/yd717_nrf24l01.c
+++ b/src/protocol/yd717_nrf24l01.c
@@ -110,6 +110,7 @@ enum {
 
 static const char * const yd717_opts[] = {
   _tr_noop("Format"),  "YD717", "Sky Wlkr", "XinXun", "Ni Hui", "SymaX4", NULL,
+  _tr_noop("USE A.ACK"),  _tr_noop("Yes"), _tr_noop("No"), NULL,
 #ifdef YD717_TELEMETRY
   _tr_noop("Telemetry"),  _tr_noop("Off"), _tr_noop("On"), NULL,
 #endif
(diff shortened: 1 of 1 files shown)


mwm:
I'm not convinced this is a good idea for a number of reasons.  First, the ack's not being received is a sign that your hardware isn't set up correctly, and I'm not sure we should make that easy to hide.

Second, I had the exact same issue with my V323, which uses the V202 protocol. Someone who's run into this "fix" on a Syma might expect the same thing to work on the V202, meaning we'll get pressure for another such kludge so people don't have to fix their broken hardware.

The more protocols this gets added to, the more likely it is that we'll run into one where the symptoms are from a different source, and disabling them won't fix things, and may indeed make things worse.

martinbudden:
Another option would be to reduce the auto retry count - it's currently 10 which seems a bit much.

hexfet:
I'm not opposed to this change.  The AA doesn't really add anything to the protocol since no data is returned from the aircraft.  I'd even support removing it completely, without the protocol option.  No reason we have to duplicate the protocol exactly.

Slightly better would be to require it for the bind phase only, so the tx won't exit the bind phase without getting a response from the aircraft.

PR #36, opened 2016-09-03
FrSky 1 way: Fixed ID and 8 channels

pascallanger:
ID is on 15 bits only not 16 bits.
The protocol is now supporting 8 channels and not only 4 has before.


victzh:
OK, as nobody else is looking, I rely on Pascal's correctness and merge it.

PR #43, opened 2016-10-01
Fix FrskyX dropout issue

hexfet:
petsmith discovered that telemetry packets may have corrupted values, including for the s.port data length value in pkt[6].  This length was used for looping over the data and processing incorrect large values caused protocol timing issues resulting in dropouts.

Added additional checks on telemetry packet to fix the issue:
- received packet length must match expected value
- crc check on packet contents
- length of s.port data must match expected value


hexfet:
Merging as I approve also :)  I've tested this change with Frysky XSR receiver.

vladislavy:
I have continuous reboot with FrSkyX protocol, at two my transmitters Devo8s + 4in1, but not at Devo12s + 4in1. All other CC2500 based protocols work fine.

hexfet:
Please connect your 8s in USB mode, save the errors.txt file, and paste here.
Then edit the tx.ini file to change the first line, current_model, to a model that is not set for FrskyX protocol.
Turn the tx off and on to boot normally, go to the model using FrskyX, and change the Fixed ID before trying to bind.

This worked for me, though I'm surprised it happened to someone else.  My errors.txt file indicates the get_tx_id function gets stuck in an infinite loop, but haven't looked into it yet.  At first glance don't see how it could happen.

vladislavy:
It's definitely "dead loop" somewhere at while loops:

static void initialize(int bind)
{
CLOCK_StopTimer();

// initialize statics since 7e modules don't initialize
fine = Model.proto_opts[PROTO_OPTS_FREQFINE];
fixed_id = (u16) get_tx_id();
failsafe_count = 0;
chan_offset = 0;
FS_flag = 0;
channr = 0;
chanskip = 0;
ctr = 0;
seq_last_sent = 0;
seq_last_rcvd = 8;

while (!chanskip)
chanskip = (get_tx_id() & 0xfefefefe) % 47;
while((chanskip - ctr) % 4)
ctr = (ctr + 1) % 4;
counter_rst = (chanskip - ctr) >> 2;
...

If I comment out both "while" transmitter boot normally.

hexfet:
Thanks!  I'd never looked closely at that code before. The first loop will never terminate if get_tx_id returns 0, or with one bits only in the last bit of each nibble, or a multiple of 47 that's unchanged by the and operation.

get_tx_id always returns the same value for a given transmitter and fixed ID.  A random number would work - maybe seed the lfsr from the clock?  I'll give it a little thought.

PR #44, opened 2016-10-02
Add Assan protocol

goebish:
Add Assan protocol for nrf24l01.
Ported from DIY-Multiprotocol-TX-Module
Forums thread: http://www.deviationtx.com/forum/protocol-development/2038-what-about-assan


src/protocol/assan_nrf24l01.c
```diff
@@ -0,0 +1,311 @@
...
+{
+    switch (state)
+    {
+    // Bind
+        case BIND0:
+            //Config RX @1M
+            NRF24L01_WriteReg(NRF24L01_05_RF_CH, RF_BIND_CHANNEL);
+            NRF24L01_SetBitrate(NRF24L01_BR_1M);                    // 1Mbps
+            NRF24L01_SetTxRxMode(RX_EN);
+            state++;
```
victzh: It's a bit unclear, can we use explicit state = BIND1? It's probably justified in the case of DATA2, DATA3, but I'd avoid it as non-local - you need to check enum to figure what follows what. Especially in the transition from BIND to DATA.

src/protocol/assan_nrf24l01.c
```diff
@@ -0,0 +1,311 @@
...
+            if(++packet_count == 217)
+                state &= ~WAIT;
+            return 10000;
+    // Normal operation
+        case DATA0:
+            // Bind Done
+            PROTOCOL_SetBindState(0);
+            NRF24L01_SetBitrate(NRF24L01_BR_250K);                  // 250Kbps
+            NRF24L01_SetTxRxMode(TXRX_OFF);
+            NRF24L01_SetTxRxMode(TX_EN);
```
victzh: It is customary to add /* FALLTHROUGH */ if your case does not end with a break to notify others that you meant it.

src/protocol/assan_nrf24l01.c
```diff
@@ -0,0 +1,311 @@
...
+                    packet[15]=0x99;
+                    for(u8 i=0;i<4;i++)
+                        packet[16+i]=packet[23-i];
+                    packet_count=0;
+                    return 10000;
+                }
+            }
+            return 1000;
+        case BIND2|WAIT:
+            if(++packet_count == 27) // Wait 270ms in total...
```
victzh: Why this arcane logic? Why not just return 270000 from line 174?

src/protocol/assan_nrf24l01.c
```diff
@@ -0,0 +1,311 @@
...
+            NRF24L01_WritePayload(packet, PACKET_SIZE);
+            if(packet_count==20)
+            {
+                state++;
+                state |= WAIT;
+                packet_count = 0;
+            }
+            return 22520;
+        case DATA0|WAIT:
+            if(++packet_count == 217)
```
victzh: Also here, just return 217000+22520 from line 195.

src/protocol/assan_nrf24l01.c
goebish: That's because the callback returns an uint16_t

src/protocol/assan_nrf24l01.c
goebish: That's because the callback returns an uint16_t

src/protocol/assan_nrf24l01.c
goebish: Done.

src/protocol/assan_nrf24l01.c
goebish: Done.

src/protocol/assan_nrf24l01.c
victzh: You're right. May be we should think about extending it to uint32_t.

PR #51, opened 2016-10-17
Update FrskyX protocol to echo S.Port telemetry to trainer port

hexfet:
Echoes each raw telemetry packet to trainer port.  If PPM In is being used the sending is disabled. The S.Port CRC is calculated and added as this value is not received over-the-air. 

The bit rate is 57600 for compatibility with devices designed to read s.port telemetry directly.  The trainer port output would need to be inverted for input to those devices.

If PPM In is in use, then changed to None, the S.Port output will not start until the tx is power-cycled.


src/target/at9/at9_uart.c
```diff
@@ -15,35 +15,13 @@
...
-
-    /* Finally enable the USART. */
-    usart_enable(USART1);
-}
-
-void UART_Stop()
-{
-    usart_disable(USART1);
-}
+#include "target/common/devo/uart.c"
```
victzh: Is it the best way? May be better to delete this file altogether and include target/common/devo/uart.c into Makefile sources?

src/universaltx/utx_uart.c
```diff
@@ -21,29 +21,5 @@
...
-
-    /* Finally enable the USART. */
-    usart_enable(USART4);
-}
-
-void UART_Stop()
-{
-    usart_disable(USART4);
-}
+#include "target/common/devo/uart.c"
```
victzh: Also here, see my comment for at9_uart.c

src/target/at9/at9_uart.c
hexfet: I did it this way because there were already target specific uart files for at9 and utx, maybe because PB anticipated specialization needed?  Agree with your suggestion though and will change.

hexfet:
To address victzh's comments the following changes are made:
- Add defines to fully isolate UART_* functions from physical USART port
- change at9 target to use  common/devo/uart.c
- revert changes to universaltx and x9d files.  These are not part of standard build and not clear UART_* changes are applicable to these projects.

Also contains bug fix for swapped hour and minute values in GPS hub telemetry.

PR #59, opened 2016-11-09
Add LQI and LRSSI telemetry values to Frsky and FrskyX

hexfet:
The LQI (Link Quality Indicator) and LRSSI (Local RSSI) indicate the quality and signal strength of the telemetry signal from the receiver.  The LQI can be used as a guide for adjusting the fine frequency protocol option.  Lower LQI is better and values under 50 are typical.  The LRSSI units is (approximately) dBm.

Above paragraph is from the manual entry (which I already updated thinking I'd added this to PR 51 - can revert if not accepted).


src/protocol/frsky2way_cc2500.c
```diff
@@ -297,7 +297,14 @@ static void frsky2way_parse_telem(u8 *pkt, int len)
     Telemetry.value[TELEM_FRSKY_RSSI] = pkt[5]; 	// Value in Db
     TELEMETRY_SetUpdated(TELEM_FRSKY_RSSI);
 
+    Telemetry.value[TELEM_FRSKY_LQI] = pkt[len-1] & 0x7f;
+    TELEMETRY_SetUpdated(TELEM_FRSKY_LQI);
+
+    Telemetry.value[TELEM_FRSKY_LRSSI] = (s8)pkt[len-2] / 2 - 70; 	// Value in Db
```
hexfet: Would it be better to also convert the regular RSSI value to (approximate) dBm?  Some people may already have alarms set for current values.

vladislavy:
I think "break" after "return" will never be used:
case TELEM_FRSKY_LRSSI:     return 255; break;

hexfet:
Thanks!  Code is updated.

PR #60, opened 2016-11-27
Bayang telemetry for h8 opensource firmware

silver13:
This adds telemetry for the bayang protocol , which works with the opensource firmware at https://github.com/silver13/h8mini-testing

The protocol works the normal , original way when telemetry is not enabled.

When telemetry is enabled, the protocol switches to rx and waits for telemetry after the packet is sent.

The code polls at 1000 but sends every second call when telemetry is off, to maintain original timing with telemetry off.. With telemetry on it sends with a 5000 period to allow for reception time. Due to this timing differences, a different bind frame is sent when telemetry is enabled, to prevent the tx from binding to a quad without telemetry, and work at possibly degraded performance.

Telemetry values:
I used the DSM type, currently 2 voltages and quad reception rate / tx telemetry rate are implemented.

Some guys in the H8 thread have already tried this code on devo 7e and devo10, and they had positive feedback. The code size difference is approx +800 bytes

Original post where I tried the protocol - might have additional info
https://www.rcgroups.com/forums/showpost.php?p=36279599&postcount=6575


silver13:
without modifying it , the changes are not saved correctly when main page items are changed to display telemetry. More precisely, they are not loaded after a power down.

goebish:
Ah ok I thought it was not necessary, I've to check other protocols then ...

silver13:
yes, at least one ( clie flier) also does not save

goebish:
Ok, and that's my fault because I told the author it was not necessary ... 
I'll handle that.

silver13:
I updated the code after running indent

no actual code changes were made

silver13:
Excellent

PR #68, opened 2016-12-06
Fix warning: comparison between signed and unsigned

vladislavy:


vladislavy:
Ooops, I use temp branches for different fixes, but anyway they automatically merged.

PR #79, opened 2016-12-18
Add support for 4-in-1 SPI switch.

victzh:
This is support for 4-in-1 flash SPI switch module. The module is Banggood's 4-in-1 RF bare module http://www.banggood.com/DIY-2_4G-CC2500-NRF24L01-A7105-CYRF6936-Multi-RF-4-IN-1-Wireless-Module-p-1046308.html and add on "motherboard" which is plug compatible with Devo 10 connector. The module has an additional SPI flash to provide Devo F-series and Radiolink AT9 with normal FAT filesystem.


goebish:
oops, looks like the linker is not happy with that.
https://travis-ci.org/DeviationTX/deviation/builds/185102080

victzh:
Yes, it's not in check in builds, so pull request is OK, but nightly build is not. I will fix it today.

PR #95, opened 2017-01-08
Devo F7 input

xuyuan:
support HOLD/TRN and AUX2, tested with my devo F7


goebish:
Seems this has broken the f4 build:
https://travis-ci.org/DeviationTX/deviation/builds/191880461

goebish:
Proposed a fix:
https://github.com/DeviationTX/deviation/pull/100

PR #98, opened 2017-01-12
faster byte reverse for xn297 routines

silver13:
This is a faster bit_reverse(x) routine using the "rev" asm instruction.

The rev instruction is available on cortex m3, and some other , but not on m0 for example.

I used a define to exclude it in the emulator build, the targets I have tried all build ( 7e , 8 , 12 , 12e and at7 ) and if this instruction is not available the compiler will throw an error so it's pretty easy to fix.

I have tested this with bayang protocol in the 7e version


silver13:
I guess using #if ( __ARM_ARCH_ISA_THUMB==2) would detect cortex m3 and higher

I found this site which lists predefined macros for cortex:
http://micro-os-plus.sourceforge.net/wiki/Predefined_macros#ARM_Cortex-M4

I'll update the code later to use it, it does seem to work in a quick check

silver13:
Should be included only on arm m3 or higher now , and gcc compiler , I have no way to rest on anything else apart from ARM however

PR #102, opened 2017-01-18
Use known good settings for F12E video.

victzh:
1.3a settings work well for me. To test it for a wider audience we need to have nightly build available.


vladislavy:
I think we can use "tw8816_init_1.3.h" with one correction only:
-0x70,  0x60,
+0x70,  0x20,
it will be equal to "tw8816_init_bl1.3a.h" with skipped registers 0x9e - 0xAE.

PR #105, opened 2017-01-18
Filesystem for the emu_devo12e

F-D-R:


vladislavy:
I think will be better (in order to get "filesystem/devo12e", not "filesystem/devo10"):

**Makefile.inc**

> SCREENSIZE  := 128x64x1
> FILESYSTEMS := common base_fonts 128x64x1
> FONTS        = filesystem/$(FILESYSTEM)/media/12normal.fon \
>                filesystem/$(FILESYSTEM)/media/04b03.fon
> LANGUAGE    := devo10
> 
> include target/common/emu/Makefile.inc
> 

**emu.h**

> #ifndef _EMU_H_
> #define _EMU_H_
> 
> #define ZOOM_X 2
> #define ZOOM_Y 2
> #define EMU_STRING "DEVO-12e-Emu"
> #endif

F-D-R:
The other transmitters share the filesystem too, like for example the DEVO 6 uses the filesystem of the DEVO 8, because they are the same, just like the one of the 12E is the same as of the 10...

vladislavy:
It was in a past. Now all transmitters use own filesystems, even they are the same.

F-D-R:
I wasn't talking about wether the built zip includes it or not, but about the naming: the emu_devo6 build still contains the filesystem/devo8 folder, because they share the same filesystem content.
Did I understand you right?

F-D-R:
I've compared to your suggestion, and found only one difference, that I haven't removed this line:
FILESYSTEM  := devo10
Is that what causes that the emu filesystem is named devo10, isn't it?

vladislavy:
**AND**:
emu.h
#define FILESYSTEM_DIR "filesystem/devo10"

emu_devo6 has "filesystem\devo6" folder. From a moment PB decided to have personal filesystems for all transmitters and emulators. Devo 12e have remainders from v5.0.0.

F-D-R:
Yeah, I hasn't reached to the code yet., because the emulator didn't work at all without the filesystem.  :)
I will examine it further, but you can push those changes as well, because I can only do that in the evening...

PR #115, opened 2017-01-24
Update "src/gui/480x272x16/_gui.c"

vladislavy:
"src/gui/320x240x16/_gui.c" and "src/gui/480x272x16/_gui.c" now are equal.

...
void _gui_draw_background(int x, int y, int w, int h)
{
	LCD_DrawWindowedImageFromFile(x, y, "media/backgrnd" IMG_EXT, w, h, x, y);
}
...


F-D-R:
I wouldn't like to merge this, because I've just made some changes (pill request #117 ), where I made this file to differ from the 320x240 equivalent: now it has the menu button with different size.

vladislavy:
No problem, it change nothing.

F-D-R:
I've changed the menuitems from buttons to labels, so there is no more need for different 320x240 and 480x272 code, so I applied your change and simply included the 320x240 variant in #156.

PR #128, opened 2017-01-31
Increase maximum number of model files from 100 to 255

hexfet:


vladislavy:
Transmitter reboot with large number of models. With >150 models transmitter reboot with any attempt to load new model. With lower number of models transmitter reboot at attempt to use selected model.

PR #142, opened 2017-02-08
Use fexists() to check model files

vladislavy:


PhracturedBlue:
Does fexists even exist in the filesystem code?  I don't remember implementing it for devofs and it may be disabled for petitfs.  What is the point  of this change?

vladislavy:
misc.c 

```
int fexists(const char *file)
{
   FILE *fh = fopen(file, "r");
   if(! fh)
       return 0;
   fclose(fh);
   return 1;
}

```
It used several times in code. I use it to save some space for devo7e.

PhracturedBlue:
thanks for the response, looks good

PR #143, opened 2017-02-08
Removed the unused fontColor from the button calls

F-D-R:
It spared some space too, the 7E image went from 113.71kB to 113.64kB.

### DIFF
--- a/src/gui/button.c
+++ b/src/gui/button.c
@@ -25,7 +25,7 @@ enum {
 };
 
 guiObject_t *GUI_CreateButton(guiButton_t *button, u16 x, u16 y, enum ButtonType type,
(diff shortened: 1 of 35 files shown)


vladislavy:
I see one GUI_CreateButtonPlateText with white (0xffff) font color. If you remove font color what will be with white font color?

F-D-R:
No, that value was never used.
Inside the function it was assigned to a member of the button struct, but when it actually draws the button, it uses the color member of the font, which is an other member of the button.

F-D-R:
I would like to clear, that this change doesn't disable setting the button's font color, because that comes from the font definition in media/config.ini
These parameters were obsolete...

PR #144, opened 2017-02-09
Fix change model icon

vladislavy:
With "u16 total_items" devo10 emulator crashed if I try to change default model icon. With "u8 total_items" it is OK.

It was changed from "u8" to "u16" with "Pull Request #128: Increase maximum number of model files from 100 to 255"


F-D-R:
I found this bug too: if the first list you open is the icon list, then emulator crashes (any of them), but if you opened an other list first, for example the model load, and only after that the icon list, then it works fine.
The TX fw doesn't have this problem.

PR #148, opened 2017-02-11
Move button's font definition into _DrawButton() method

vladislavy:


vladislavy:
I think we need to synchronize all "config.ini" files and use same font names for all used GUI. For example, 128x64x1 "config.ini" has not [font-button] section, therefore BUTTON_FONT doesn't work normally for buttons. I think that all "config.ini" files must contain all font sections, even if font not used right now. It will useful for future changes.

F-D-R:
I agree.
I would like more customization, that's why I introduced the [font-menu] too, but I'm not yet finished...

BTW do you remember how did we make the fonts? I'm not satisfied with the ones we've got.
For example the bold font, which is used for the model name on the color transmitters doesn't have the special characters, and I don't like the font used in the BW transmitters at all...

vladislavy:
Look to Makefile:
######################
#Necessary Font files#
######################

I don't know can these fonts be used:
https://slackbuilds.org/repository/12.2/system/efont-unicode-bdf/

PR #152, opened 2017-02-12
Make production type default for the emulator build

vladislavy:


F-D-R:
What's the difference?
Is it only the debuglog screen in the menu?

vladislavy:
Yes, I'm tired to add TYPE=prd to emulator build command line. The main build has long time TYPE=prd as default.

TheRealMoeder:
I disagree. I forgot example use buildtype_dev to display some debug information only on developer builds and emulators. What's the disadvantage of keeping  emulator builds dev by default?

vladislavy:
I don't know how it will be with Linux, but with Windows if I press menu item "Debuglog" the emulator will crash. I can build the emulator developer version with TAPE=dev, but why I need developer version every time? Or every time add TAPE=prd for normal version?

TheRealMoeder:
Then we should look into fixing the problem in the windows emulator. The Linux emulator works fine for me on upstream code.

PhracturedBlue:
I don't really see the point either  The primary purpose of the emulator is to debug stuff anyway.Sure I guess a couple of users may use it to test complex rules (and some new users may want to try the GUI before flashing), but that has to be the minority compared to developers using it to verify that new code works as expected.

vladislavy:
I can't find point of emulator "Debuglog" problem at main branch, but for BRANCH_5.0 it started after Jan 25, 2017, "Reorganize filesystem build, from Geoffrey Hausheer". It change almost nothing except Makefiles. It looks like emulator don't know way to debuglog data.

PR #154, opened 2017-02-18
Add "Ultimate Devo F12E" (devof12e-4Mb)

vladislavy:
Ultimate Devo F12E is modified Devo F12E with replaced flash memory chip UPCT25VF512A (512 Kbit) to SST25VF032B (32 Mbit). It has not devofs and use FAT filesystem (USB drive).


TheRealMoeder:
Looking good...

F-D-R:
I think the name is confusing.
4Mb is 4 megabits, but AFAIK the memory size is 4 megabytes, i.e. 4MB.

Furthermore I'm not sure that all HW change needs an own firmware. I know that the ultimate 7E has got its own too, but what if the users of F7 and F4 start to replace the flash chip too? :)

Otherwise it's a good job!

F-D-R:
Did you closed this because you will rename it, or because of the nonworking standard GUI?

PR #171, opened 2017-02-23
Make DR&Expo work again on F12E

TheRealMoeder:
Layout still messed up!


vladislavy:
Thank you very match, now I understand how it work better. But, I think it is not fully correctly since "value1" and "value2" placed at one line, one by one. So I will use other values.

PR #174, opened 2017-02-23
Revert "Fix "text" GUI "Standard" - DR can't be greater 100%"

TheRealMoeder:
You are too fast! This doesn't solve the problem, actually it is worse than before: DR above 100 % can make sense, you now scale the rates differently so when selecting 125 % you actually have 100 %. That is not correct. The problem lies in the XYgraph indeed, which can be seen in the Devo10 emulator which also draws incorrectly above 100%, but at least doesn't crash the whole thing. I suggest reverting the commit.

### DIFF
--- a/src/pages/128x64x1/standard/drexp_page.c
+++ b/src/pages/128x64x1/standard/drexp_page.c
@@ -91,9 +91,8 @@ void PAGE_DrExpInit(int page)
                      2 * LINE_SPACE, count, row_cb, NULL, NULL, NULL);
 
     GUI_CreateXYGraph(&gui->graph, GRAPH_X, GRAPH_Y, GRAPH_W, GRAPH_H,
-                      CHAN_MIN_VALUE, CHAN_MIN_VALUE * 5 / 4, 
-                      CHAN_MAX_VALUE, CHAN_MAX_VALUE * 5 / 4,
-                      0, 0, show_curve_cb, curpos_cb, NULL, NULL);
+            CHAN_MIN_VALUE, CHAN_MIN_VALUE, CHAN_MAX_VALUE, CHAN_MAX_VALUE,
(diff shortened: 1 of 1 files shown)


TheRealMoeder:
I just checked....the internal scaling is correct, but the graph is not. Anyways, not the real solution to the problem. The bug is even more pronounced on devo10 emulator when you scale to 125% and then move the input to max or minimum, which makes a black vertikal line appear...

vladislavy:
Advanced mixer have the same solution. I don't invent nothing new.

TheRealMoeder:
True, but nevertheless wrong. I will see if I can spare voice alert time to look into it :)

vladislavy:
From the beginning mixers have limit 100% and you can set up to 125% at channels properties. Then things were messed up by ability to set mixers up to 125%.

F-D-R:
True...  :)

TheRealMoeder:
The question is what do we want? The current (to be honest never before questioned) solution has different scales on X and Y axis, which maximizes screen usage (which is good!). On the other hand the curves would probably look more to what you expect if they had the same scale on X and Y axis and we just don't show any line beyond the Channel input limits.

F-D-R:
I don't know, the graph might be bad on the BW screens.
On the color screens it has a longer Y axis, because the Y can go up t +-125%, while the X axis is limited to the +-100% input range.
I thought that you question the defaul limited +-100% output range, which is not the whole throw, to allow the users to adjust it wider to fit the needed servo throw.
It would break the backward compatibility if we changed that.
On the other hand the graph is relatively easy to change...

TheRealMoeder:
No, I don't want to change any of the handling, I was just wondering about the graph itself. They way it is done in the color screens seems better to me. If we have vertical screen space left on the pages it is needed, why not stick with the same solution of extending the y axis? But I admit this is not important and I'm fine with keeping things the way they are

PR #177, opened 2017-02-23
Add GW008 protocol

goebish:
Add Global Drone GW008 protocol.

Channel 6: flip

There are 3 versions of this small quad, this protocol is for the one with a XNS104 IC in the stock Tx and PAN159CY IC in the quad (SOCs with built-in xn297 compatible RF).
The xn297 version is compatible with the CX10 protocol (green pcb).
The LT8910 version is not supported yet.

This PR also includes enhanced packet mode support for xn297 emulation (tx & rx).

Forum thread: https://www.deviationtx.com/forum/protocol-development/6420-global-drone-skull-drone-or-gw008


TheRealMoeder:
If there are three versions of the drone, do you think it is clever to just name this protocol GW008?

goebish:
Any suggestion to rename it ? Or just add a useless "XNS104" format option ?
I think that's not a problem as long as it's stated in the model compatibility list that this protocol is only compatible with this particular model.
But I admit it's starting to be a mess, look at the CX10 for example, it has 5 (!) different protocols, and only 2 of them are named CX10.

PR #179, opened 2017-02-24
Fix 128x64x1 main_page.c

vladislavy:


vladislavy:
X, Y, width, height can't be numerical, we need to use macro variables because of "test" GUI.

F-D-R:
Do you mean "text" GUI?

vladislavy:
Sorry, yes it is "text" GUI.

PR #182, opened 2017-02-25
Fix change icon for Devo F12E

vladislavy:


vladislavy:
With the second fix we can use BW *.pbm images from any editor, without editing the file header.

PR #194, opened 2017-03-02
Fix "text" GUI button and textsel

vladislavy:


vladislavy:
I'm not 100% sure it is right. First, I fix these lines, but it doesn't look nice with Devo F12E GUI, so I commented these lines out. Now Devo F7 GUI look nice too. I will delete these lines.

PR #201, opened 2017-03-03
Adding LISTBOX_FONT, and eliminating the global labelDesc font

F-D-R:


vladislavy:
FDR, even before this commit: if I visit "Telemetry monitor" pages (both pages or two or several times first page), next I try to change model icon, default icon disappeared or not, I walk on icons and emulator eventually crashed. I can't find the reason (emu_devof12e and emu_devo10).

F-D-R:
I don't think it has anything to do with these changes.
I went back to pull request #100 in 14th of january, and it was already wrong then.
I will look for the cause further. The official 5.0.0 hasn't got this but, but back then there was no list on the icon selection, but a textselect field...

vladislavy:
It started after replace Listbox with Scrollable, and crashed only with default icon. If I don't touch default icon, emulator do not crash.
https://github.com/DeviationTX/deviation/commit/db1bef098aefe9a34be34a9f77e44d8570c30dad

F-D-R:
Pushed a temporary solution in #205

PR #205, opened 2017-03-04
Getting around the icon crash after telemetry page

F-D-R:
I couldn't find the bug itself, just got around it to prevent the crash.
Somehow the modeltype gets an invalid value...


vladislavy:
First, about type < 0, type always positive and we have compile warning.
Second, Default icon still disappeared, but emulator worked.
Third, emulator has slow response on Default and last icons, it seal on these icons.

And, even without fix real devof12e work without problems. It seems it is one of mysterious Windows built emulator crashes.

vladislavy:
Fixed it completely. Model Load/Save page can get dirty memory for struct model_page. In that case mp->modeltype can get any value, since mp->modeltype is not initialized at page init.

I placed the pull request.

F-D-R:
Actually the ModelType is an enum, which is basically an int, so it can have negative value.
Of course it was only a workaround, and the final solution would be to not letting it have an invalid value, just like you did...
The valid range check does no harm, but the text based transmitters might not have the noicon anyway.
I will push again the telemetry page font changes though...

PR #206, opened 2017-03-05
Fix emulator reboot on Load/Save page

vladislavy:


vladislavy:
Model Load/Save page can get dirty memory for struct model_page. In that case mp->modeltype can get any value, since mp->modeltype is not initialized at page init.

PR #208, opened 2017-03-06
Fix emulator crash at debuglog page

vladislavy:
FDR, all lines are centered at debuglog. Can you fix it to left aligned?


F-D-R:
Aligning is no problem. Now it has the DEFAULT_FONT, which is center aligned by default, but I was thinking about if it deserves a separate list font type.

However in this bug fix you only partially initialized those arrays.
Shouldn't they be fully initialized, i.e. all the array elements?
Probably it is easier to set all those values with memset...

vladislavy:
It initialize all members of an array to the same value. Elements with missing values will be initialized to same value. The "volatile" type doesn'y allow to use memset.

F-D-R:
AFAIK all missing values of a fixed size array will be 0.
In the fitst case it is good, but would be the same without the initializer too, but only the first item of the line_pos will be 0xFF...

PR #211, opened 2017-03-07
Fix debuglog page

vladislavy:
1. Fix Windows built emulator crash at debuglog page
2. Fix debuglog circularity at debuglog page. Debuglog entries repeated 2, 3, 4... times at debuglog screen, depending from length of entries.

Now it is clear that the size of the debuglog is too small. It can contain very limited number of entries.
In my case it can contain 15 entries: "Backlight: X", thats all. 13 x 15 = 195 and debuglog size = 200.

### DIFF
--- a/src/pages/common/_debuglog_page.c
+++ b/src/pages/common/_debuglog_page.c
@@ -15,14 +15,14 @@
 
 ctassert((DEBUG_WINDOW_SIZE < 255), debug_window_size_too_big);
 #define NUM_ROWS 20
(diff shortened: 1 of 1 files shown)


vladislavy:
About debuglog size? I think more useful will be about 400 - 500. But I don't know why now it limited to 255. ctassert((DEBUG_WINDOW_SIZE < 255), debug_window_size_too_big);
May be it is devo7e limitation?

vladislavy:
OK, we can decrease NUM_ROWS 20 in order to not have empty lines with numbers on the screen. Why we need scroll for debuglog with 4-5 entries. Most debug messages are quite long.

Or do nothing and leave it as is.

PR #212, opened 2017-03-07
Aligning the DebugLog left by using the LABEL_FONT

F-D-R:


F-D-R:
I will rather create a new font type for the read only lists...

PR #218, opened 2017-03-17
Introducing the vastly improved voice alert and voice menu feature

TheRealMoeder:
based on the initial work by MWM and phantom-8


src/extended_audio.c
```diff
@@ -0,0 +1,211 @@
...
+#include <libopencm3/stm32/usart.h>
+#endif // EMULATOR
+
+#if HAS_EXTENDED_AUDIO
+
+static u32 audio_queue_time = 0;
+
+// Initialize UART for extended-audio
+void AUDIO_Init() {
+#ifdef BUILDTYPE_DEV
```
victzh: You can omit #ifdef BUILDTYPE_DEV - in non dev builds printf is an empty macros so it's totally eliminated and is not present in the final binary at all.

TheRealMoeder: I thought it was not eliminated, since many protocols use a dbgprintf which is redefined...

TheRealMoeder: Just rechecked std.h, you are of course right...I will remove the ifdefs

TheRealMoeder:
what do you suggest?

PR #223, opened 2017-03-21
Add Fayee FY805 format to MT99xx protocol

goebish:
Add Fayee FY805 format to MT99xx protocol, ported from MULTI-Module.

Channel 6: flips
Channel 9: headless

Forums thread: https://www.deviationtx.com/forum/protocol-development/5329-fayee-fy805-captures


TheRealMoeder:
Maybe update deviation-manual as well...

goebish:
Yes I'll update the manual and wiki as well, they're a few protocols late.

PR #245, opened 2017-03-27
Fix GUI_DrawXYGraph()

vladislavy:
GUI_DrawXYGraph() has miscalculation for point and axis position.


vladislavy:
Incorrect width calculation for numerator. Calculation for denominator is right.

PR #250, opened 2017-03-28
Add Saving music for color GUI

TheRealMoeder:


TheRealMoeder:
It's played whenever you need to press save (b&w)  /ok(color) or else the changes will not be saved, in mixer setup. Changing for example transmitter configuration  gets always saved when exiting the menu, so we have to differentiate. Either this or we completely remove play music saving for all screens

F-D-R:
But still, those changes are not yet saved then.
I don't see too much value in that, because the problem usually is that some people doesn't recognize there is a Save/OK button at all, so they exit without saving...

TheRealMoeder:
I agree, shouldn't we then do the opposite and remove MUSIC_SAVING from B&W screens as well?

F-D-R:
If you ask me, yes we should, just to be consistent across the whole Devo line...

TheRealMoeder:
Agreed.

PR #252, opened 2017-03-31
cflie_nrf24l01 protocol support for CPPM CRTP commander type

theseankelly:
Adding support for a new packet type for the Crazyflie 2.0 -- CPPM emulation via the NRF radio. Allows for flexible channel support and re-mapping. Eliminates need to hardcode max angle/rates in the deviation protocol code.

### DIFF
--- a/doc/CFlie_NRF24L01.txt
+++ b/doc/CFlie_NRF24L01.txt
@@ -1,0 +1,54 @@
+Crazyflie/Crazyflie 2.0 Protocol (uses NRF24L01) 2017 April 1
+
+# Description #
+This protocol is for supporting the Crazyflie and Crazyflie 2.0 nanocopter.
+Comprehensive details about this copter can be found on the crazyflie wiki and/or forum:
+wiki: http://wiki.bitcraze.io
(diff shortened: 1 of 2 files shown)


theseankelly:
Definitely will scrub tabs -- I never use tabs on purpose unless the coding convention calls for them. My mistake. 

Will add documentation as well

theseankelly:
Docs added, found one tab and removed it.

F-D-R:
I've found a lot more in this block:
`static u8 toc_size;				// Size of the TOC read from the crazyflie
static u8 next_toc_variable;	// State variable keeping track of the next var to read
static u8 vbat_var_id;			// ID of the vbatMV variable
static u8 extvbat_var_id;		// ID of the extVbatMV variable
`
...but yes, that one was which I spotted first.

theseankelly:
Gotcha. I missed that 'cuz it's from a previous commit and was using 'git diff' to find the tabs. Scrubbed the whole file for '\t' and I believe I've got them all now.

TheRealMoeder:
Although I actually meant documentation in deviation-manual. But we now it can be easily added there as well

PR #253, opened 2017-04-02
Fix compiler warning, ensure welcome message if no splash page, fix emu7e build

TheRealMoeder:


vladislavy:
There is not sense to have separate case for default, it's only for compiler do not think that possible to have vpt unassigned. At any case VOICE_SRC_MIXER will be default, but compiler don't know about it.

TheRealMoeder:
Updated pull request, I guess this is now the most reasonable and memory sparing approach ;)

TheRealMoeder:
No it won't, because voice even without menu does by far not fit on the 7e, unless we exclude standard mixer gui

F-D-R:
I see, thanks...

PR #261, opened 2017-04-19
Fix  4ch_dr.ini template

mwm:
Fix SRC for first mixer in AIL & ELE channels .


vladislavy:
Most likely it has to be:

```
[channel1]
template=expo_dr
[mixer]
src=ELE
dest=CH1
[mixer]
src=ELE
dest=CH1
switch=ELE_DR
scalar=60

[channel2]
template=expo_dr
[mixer]
src=AIL
dest=CH2
[mixer]
src=AIL
dest=CH2
switch=AIL_DR
scalar=60

```

PR #264, opened 2017-05-05
Update lang.fr

Morlacus:
dditions to the French lanquage file. May be some texts I added are too long. I do not know if there is a limit. I corrected 2 entries which were wrong and changed 'low rate' in 'petit déb' instead of 'grand déb' and 'high rate' in 'grand déb' instead of 'petit déb'. This could puzzle french users which have already the current translation.

### DIFF
--- a/src/fs/language/lang.fr
+++ b/src/fs/language/lang.fr
@@ -1,7 +1,8 @@
 ﻿Français
 :# Channels
 Nbr canaux
->:%d bytes left
+:%d bytes left
+%d bytes restant
 : is %d%%, safe value = %d%%\n
  = %d%%,valeur sécu. = %d%%\n
 : is %d%%,\nsafe value = %d%%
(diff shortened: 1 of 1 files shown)


goebish:
I'd rather use "taux haut" and "taux bas"

Morlacus:
In French we speak of "Débattements" (Déb)  for the travel   of a control surface. As concerns D/R the standard spelling are "grand", "moyen" and "petit" "débattement" The original translation were correct except an inversion between high and low which were not congruent with the english.

goebish:
Ok, I'm French but I always set my stuffs to English language, so you're probably right ;)
However, I'll check with the emulators to make sure string lengths aren't an issue.

Morlacus:
Ok thanks . There is one translation for which I am not very satisfied. It is for Toggle. I tried Basculant but there is possibly better translations. If the string is not too long, Icône basculant or Icône basc would may be better.

PR #265, opened 2017-05-06
AFHDS 2A telemetry update

goebish:
Telemetry update for Flysky/Turnigy AFHDS 2A protocol

- RSSI now reported as a positive number, allowing to set an alarm (min = 60, the lower the better)
- Report Link Quality as LQI (max = 100, the higher the better), probably a better indicator than RSSI
- External IBUS voltage sensors support, only 1 for 7e (modular build), 2 for other platforms, VOLT2 & VOLT3
- External IBUS RPM sensor support, only 1 sensor supported (not available on 7e)
- Built in external voltage sensor support (IA6C, X6B RXs...), not available on 7e
- External IBUS temperature sensor support, not tested, not available on 7e


goebish:
It can be merged now, everything is confirmed working.

PR #286, opened 2017-08-12
Branch 5.0

bartremes:
devof7-512 folder added. for devof7 with 512k memory
please merge with branch 5.0 and master so it will build with nightly builds.

for mine modified memory devof7 its is working bur please check at:

src/target/devof7-512/devof7-512_opt.ld

+	rom (rx) : ORIGIN = 0x08003000, LENGTH = 116K /*not shure if this is correct for teh 512k*/
+	ram (rwx) : ORIGIN = 0x20000000, LENGTH = 16K /*not shure if this is correct*/

thanks


F-D-R:
AFAIK the 5.0 branch isn't built with the noghtly, but Vlad maintains it and posts the builds in the forum...

vladislavy:
1. I don't see any sense to add "Devo F" series to BRANCH 5.0. Use "master" branch.
2. I don't see any flash memory type detection or indication. So, I think it works by pure accident.
3. What is the flash memory chip and how/where it's installed?
4. It's not a good idea to copy all target files to a new target. Will be better to use includes from Devo F7 target. This way you need edit one target only for all clones (similarly to Devo7e-256).

TheRealMoeder:
Shouldn't we close this  as you (vlad) are close to a PR for extra memory code for all devo f series transmitters?

PR #290, opened 2017-08-30
Add devof12e-XMS

vladislavy:
For modded Devo F12E with upgraded flash memory chip


TheRealMoeder:
did you try to fit both devofs  and fat into a single f12e build? that would actually be the most end-user friendly solution...

vladislavy:
I'm in doubt that this is possible.

PR #291, opened 2017-08-31
Add 3pos trim switch functionality

DavidBoone:
Adds the ability to use the momentary trim switches as 3 pos switches.  ie, normal resting position is centred (`0`), while pressing the trim one way or the other results in a momentary deflection in one direction (`-100`) or the other (`+100`).

I used this together with "Adjustments" in Betaflight to allow easy adjustments to my PID's.

### DIFF
--- a/src/mixer.c
+++ b/src/mixer.c
@@ -706,6 +706,18 @@ static void _trim_as_switch(unsigned flags, int i, int is_neg)
             _trim_music_play(i, is_neg, 0);
 #endif
         }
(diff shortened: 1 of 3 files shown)


F-D-R:
You could achieve this functionality before with two momentary trim switches (and two virtual channels) mixed into one channel, but of course I see that with you solution it is much easier.
Is this function used that often?

TheRealMoeder:
I agree this is not that big of a change, but makes for a lot easier setup of a trim switch as a 3way switch. And thanks for updating the manual as well!

PR #292, opened 2017-09-04
Add devof7-XMS and devof4-XMS

vladislavy:
For modded Devo F7 and Devo F4, with upgraded flash memory chip.


vladislavy:
They have minor differences. 30 models by default, datalog support, devof12e-XMS has language support and debug mode.

TheRealMoeder:
In devo7e we have also the emulator with features not available on TX (main page configuration). Can't we do the same for F-line? Datalog seems useless on emulator anyways, language can't hurt in the emulator to help configuration and all models >30 would be ignored when uploaded to TX.

vladislavy:
With devo7e it's the only way to setup main screen layout. I'm not sure that it's a good idea to add fake features to the emulator. There will be even more questions from users.

I do not have a definite solution. We can simply remove the emulators for the Devo F-XMS line and forget about it.

TheRealMoeder:
I don't think they are necessarily a bad idea, I just don't know if we really need them. I can live with having them, I just like the keep it simple idea. On the other hand, a user soldering a memory chip is probably not a novice :P. You have two approvals,  I vote "neutral", so go ahead and merge :)

PR #293, opened 2017-09-12
Fix sounds for devo12 "mymedia\sound.ini"

vladislavy:
"mymedia/sound.ini" must be processed for each function call, not just the first time.


vladislavy:
I don't understand, it has #ifdef _DEVO12_TARGET_H_

PR #294, opened 2017-09-12
Fix sounds for devo12 "mymedia\sound.ini"

vladislavy:


TheRealMoeder:
BTW, you could have just pushed the second commit to the existing branch devo12 in your repo, and the first pull request would have been automatically updated. No need to close and open a new PR.

TheRealMoeder:
Wait...doesn't it have to be:

> 
             if(fh) {
                 sprintf(filename, "mymedia/sound.ini");
                 fclose(fh);
              } else
              {
                 static char filename[] = "media/sound.ini\0\0\0"; // placeholder for longer folder name
              }
>

vladislavy:
**static** char filename[] = "media/sound.ini\0\0\0"; initialized once only, at first function call. I don't see the difference.

PR #305, opened 2017-10-13
Fix switch mapping for devo7e-256 with "nostock" hardware option

vladislavy:


vladislavy:
It doesn't hurt to anything but help me to support T8SG builds. This fix disable switch mapping for removed FMOD and HOLD switches, with enabled "nostock" hardware option.

PR #317, opened 2017-11-11
Fix S-FHSS protocol servo throw

vladislavy:
Tested up to 150%

### DIFF
--- a/src/protocol/sfhss_cc2500.c
+++ b/src/protocol/sfhss_cc2500.c
@@ -187,44 +187,42 @@ static void calc_next_chan()
 }
 
 
-// Channel values are 10-bit values between 86 and 906, 496 is the middle.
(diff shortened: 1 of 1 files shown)


goebish:
Do you have a genuine Futaba FHSS transmitter ? 
Looks like this protocol is not working well, with most receivers only the 4 first channels are working.
I've neither a Futaba TX or RX, so I can't help with that ...

vladislavy:
I have Futaba 14SG, but havn't S-FHSS receiver. User Poldi reported that Align R6GF receiver can use all 6 channels.

vladislavy:
What does it mean?

goebish:
Just someone that made a mistake, he posted an empty message referencing this PR in the other PR, I deleted it.

PR #326, opened 2017-11-23
S-FHSS fix

goebish:
- Fix for some S-FHSS receiver that didn't receive channel 4-8, now the RSFSB-B that's shipped with the T8SG has all8 channels working.
- Add failsafe support


goebish:
rcgroups + trial & error ;)

PR #327, opened 2017-11-23
Timer-based PPM protocol

hexfet:
Uses timer to generate ppm output pulses, replacing current method of bit-banging based on clock timer callbacks.  This improves control of the pulse width to reduce jitter.

The implementation is based on the timer code that had been commented-out in pwm.c, which needed fixes for defines and co-operation with usart using the same pin.  The basic logic was correct.  Some usart code is also updated for co-operation.

This has been tested with devo10 and t8sg.  Needs devo12 and modular build testing as those are slightly different.  [Test builds available](https://www.deviationtx.com/downloads-new/category/818-hexfet-ppm-timer).

### DIFF
--- a/src/protocol/ppmout.c
+++ b/src/protocol/ppmout.c
@@ -51,7 +51,6 @@ enum {
 };
 ctassert(LAST_PROTO_OPT <= NUM_PROTO_OPTS, too_many_protocol_opts);
 
-volatile u8 state;
 static void build_data_pkt()
 {
     int i;
@@ -82,6 +81,7 @@ static void initialize()
     CLOCK_StopTimer();
     if (PPMin_Mode())
         return;
+
     PWM_Initialize();
     num_channels = Model.num_channels;
     if (Model.proto_opts[CENTER_PW] == 0) {
@@ -91,9 +91,9 @@ static void initialize()
         Model.proto_opts[PERIOD_PW] = 22500;
     }
     
-    state = 0;
     CLOCK_StartTimer(1000, ppmout_cb);
 }
+
 
 const void * PPMOUT_Cmds(enum ProtoCmds cmd)
 {
--- a/src/protocol/ppmout.c
+++ b/src/protocol/ppmout.c
@@ -33,11 +33,6 @@
 static volatile u16 pulses[PPMOUT_MAX_CHANNELS+2];
 u8 num_channels;
 
-/* FIXME:  The original imlementation used a PWM to output the PPM signal.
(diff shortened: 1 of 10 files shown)


F-D-R:
I guess we should wait until somebody test this on the 7E/12 then...

hexfet:
Yes, before merging I think.  The 12 uses a different pin for ppm, and the modular build makefile had to be changed to include the libopencm3 library.  Went ahead with the PR for review and comments.

TheRealMoeder:
How exact has testing to be? My scope is not yet with me after moving, I only have the the cheap LA available, which obviously can't do jitter analysis. But if it's just a matter of checking the PPM output, I could test the 7e modular build

F-D-R:
I think that will do for now...

hexfet:
Yes, I think even just testing with the trainer input of another transmitter would be fine.

TheRealMoeder:
Please be patient, I just have not had the time to run the tests yet

OpenUAS:
@TheRealMoeder could you find some time to test @hexfet nice improvement? TIA

hexfet:
I made a modular build for my t8sg to test.  It works fine.  That leaves just the devo12 variation untested.   It would be better to go ahead and merge I think.

hexfet:
There is a problem with this implementation in a modular build. Not sure how my first test worked.  Would like opinions on the best solution.

The issue is with the ISR in target/common/devo/protocol/pwm.c.  In a modular build this code is linked into the protocol .mod file, so the address in the vector table is not correct when the code is run.

One way to fix this would be to set the address in the vector table at runtime when the ppmout protocol is loaded.  But I think this would take some post-processing to find the ISR offset in the .mod file and then patch it somewhere it could be found at runtime.  Or is there an easy way to get a function offset in the .mod file at runtime?

Another option is to move the pwm.c code into the dfu file.  The disadvantage is using more flash in the modular build.  Also I'm not sure how to make it happen in the makefiles.

Are there other options?

hexfet:
Setting the interrupt vector at runtime is not possible because the vector table is stored in flash (so much for late-night programming :/)  I've moved the pwm.c code out of the .mod file which increases the size of the modular builds by 448 bytes, moving from 115.27 to 115.71 KB.

hexfet:
Moved only the timer interrupt service routine and data pointer to rom.  Now the modular build rom memory usage is just 96 bytes more than the master branch.  Test builds are updated.  I've tested both normal and modular builds on t8sg.

PR #328, opened 2017-11-28
Add ability to specify display scale/format string

nknotts:
I wanted to be able to specify channels in ranges other than [-100, 100] %. For example, I may want to map [-100, 100] % to [-1000, 1000] RPM. I've added the ability to specify the scale factor as well as the `sprintf` format string to facilitate this. To override, specify `display-scale=xxx` and/or `display-format=sss` in a particular channel section of `model.ini`. If `display-scale` or `display-format` are not specified, the default format is used. These options are only exposed by manually editting the `model.ini` file.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -125,7 +125,8 @@ PROTO_EXTRA_OBJS := $(addprefix $(ODIR)/, $(notdir $(PROTO_EXTRA_C:.c=.o)))
 SYMBOL_FILE := $(ODIR)/symbolfile
 endif
 
-HGVERSION := $(shell ../utils/get_version.pl ${TARGET})
+HGVERSION ?= $(shell ../utils/get_version.pl ${TARGET})
+HGVERSION := $(HGVERSION)
 
 # Rebuild on hg version change.
 # Based on http://mercurial.selenic.com/wiki/VersioningWithMake
--- a/src/config/model.c
+++ b/src/config/model.c
@@ -93,6 +93,7 @@ static const char MIXER_CURVE_SMOOTH[] = "smooth";
 /* Section: Channel */
 static const char SECTION_CHANNEL[] = "channel";
 
+static const char CHAN_DISPLAY_SCALE[] = "display-scale";
 static const char CHAN_LIMIT_REVERSE[] = "reverse";
 static const char CHAN_LIMIT_SAFETYSW[] = "safetysw";
 static const char CHAN_LIMIT_SAFETYVAL[] = "safetyval";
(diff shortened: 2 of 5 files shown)


F-D-R:
Will this integer division be able to provide the required output?:
`sprintf(tempstring, val_format, val_raw/val_scale);`

To be honest I don't know the range used in `val_raw`, but I can imagine cases, when a less then one divider is needed, or when simply the integer divider steps cause too large change on the result, i.e. lack of precision...

nknotts:
I discovered via inspection of the code that the normal range for `val_raw` was [-10,000 10,000]. 

This can be seen by [src/pages/common/_main_page.c#L86](https://github.com/DeviationTX/deviation/blob/be57e8476f8a929976069fdec97d2c1b80ce23b9/src/pages/common/_main_page.c#L86)

```sprintf(tempstring, "%3d%%", RANGE_TO_PCT(MIXER_GetChannel(idx - (NUM_RTC + NUM_TIMERS + NUM_TELEM + 1), APPLY_SAFETY | APPLY_SCALAR)));```

and

[/src/mixer.h#L13](https://github.com/DeviationTX/deviation/blob/d88c192bace0bdee6e84cd0c1d20483c7c996f09src/mixer.h#L13)

```
//MAX = 10000
//MIN = -10000
#define CHAN_MULTIPLIER 100
#define PCT_TO_RANGE(x) ((x) * CHAN_MULTIPLIER)
#define RANGE_TO_PCT(x) ((x) / CHAN_MULTIPLIER)
```

My intention was to make this change as minimally invasive as possible. I considered making the division a floating point operation but was unsure of how to measure the performance impact of doing so. For the ranges I desired, decimal devision worked well. 

If floating point math is out, I could add a comment or two to help clarify.

F-D-R:
I see, thanks.
I can't say if a floating point operation could cause performance issues (I doubt), but you point about the minimalist approach is valid. I just want to be sure it is future proof enough...

F-D-R:
Let's wait for what the others think about this... :)

F-D-R:
AFAIK we don't have a documentation about the model.ini format, only MWM wrote a tutorial about that (http://rc.mired.org/p/writing-deviationtx-model.html).
However the manual indeed has a section about the only feature which has no GUI implementation, the safety system: https://www.deviationtx.com/manuals/html-devo8-en/ch_mainpage.html#safety-system

vladislavy:
It doesn't work for virtual channels values at main screen boxes. See issues.

PR #342, opened 2018-01-18
model.c: fix duplicate de/serialization of 'display-scale'

nknotts:
https://www.deviationtx.com/forum/builds/7571-strange-behavior-of-devo-10

### DIFF
--- a/src/config/model.c
+++ b/src/config/model.c
@@ -767,12 +767,6 @@ int assign_int(void* ptr, const struct struct_map *map, int map_size)
             strcpy(m->limits[idx].displayformat, value);
             return 1;
         }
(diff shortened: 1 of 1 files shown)


nknotts:
All entries in [_seclimit](https://github.com/DeviationTX/deviation/blob/a8cd2510552b6c3b6cc2eda98469a07027656607/src/config/model.c#L543) are de/serialized via the [write_int](https://github.com/DeviationTX/deviation/blob/a8cd2510552b6c3b6cc2eda98469a07027656607/src/config/model.c#L1086)/[assign_int](https://github.com/DeviationTX/deviation/blob/master/src/config/model.c#L579) functions. To prevent the duplications, I removed the additional de/serialization code for 'display-scale'.

PR #349, opened 2018-01-30
Fix Makefile for Windows environment

vladislavy:


vladislavy:
If it will be "release_$1: zip_$1 zip_win_$1 fs_$1" then you will try to build "devo8" emu but not "emu_devo8". See "both_$1: $1 emu_$1".

PR #364, opened 2018-02-19
Modify and cleanup devof12e "lcd.c"

vladislavy:


vladislavy:
I started from code cleanup and ended with modification. The original version has two different types of status check for different stages. I changed it to uniform and deleted unnecessary functions. Also, the I2C1_ReadBufferDMA() use DMA to read several bytes to buffer, I changed it to same way as for write (using i2c read) and deleted unnecessary functions and headers. Now it takes about 100 bytes less memory. Tested by about half a hour with menu navigation, scrolling icons and use video PAL and NTSC: = all OK, no errors at debug log.

PR #368, opened 2018-02-22
Add selectable and configurable drawn background

vladislavy:
You can select to draw background instead a background image for color GUI. You can set header color and background color. The settings are in "config.ini" file. By default used background image.

```
[background]
  drawn_background=0
  bg_color=50A0D8
  hd_color=707070
```


F-D-R:
Why did you only include the drawing settings for the 480x320 type screens?
How would it work for the Devo8?

vladislavy:
It's for both color screens 480x272x16 or 320x240x16
#if (LCD_WIDTH == 480) || (LCD_WIDTH == 320)

F-D-R:
Oops, sorry, I’ve missed that both values are widths... 😁
Don’t we have a switch for the color screens in general?

PR #391, opened 2018-03-23
Add devo12-XMS target

vladislavy:


F-D-R:
Couldn't we just detect the flash size and work according to the result?

vladislavy:
In that case all devo12 specific code need to be included to common code, and other significant changes. I don't think it will be worth the result.

F-D-R:
No, I thought in the DEVO 12 specific code, since it has specific memory management anyway, hasn't it?

vladislavy:
I don't understand. devo12-XMS use common code and devo12 replace some common code and add some specific code for media drive, at build time.

PR #396, opened 2018-04-18
Update UART_ API to support serial protocol development

hexfet:
Support for SBUS requires non-standard data rate and parity and stop bit changes.  Other serial protocols need similar control.


hexfet:
Considered creating a separate uart.h file instead of putting the enums in target.h, but couldn't find a good way to include it at the target.h level.  Would that be better?

PR #397, opened 2018-04-21
Update UART API to support receiving data

hexfet:
Adds UART_StartReceive and UART_StopReceive functions to UART API.  UART_StartReceive sets callback function that receives incoming serial data.
Adds UART_SetDuplex to set half duplex hardware support.
Adds TESTSER protocol as example implementation and test harness for UART_* functions.


F-D-R:
I would suggest not to commit it into the trunk, but provide as a test build...

hexfet:
I can provide a test build, but not sure why you suggest these changes should not be committed.  This is a prerequisite for supporting CRSF, not throwaway code.  But it's not dependent on specific serial protocol so should be a separate PR.  Should've mentioned that I've tested on both t8sg and devo10.

F-D-R:
Then why did you name it test serial?

hexfet:
Since TESTRF is for testing RF chips it seemed TESTSER would be good for testing the serial port.  Would it be better to only include it in dev builds like TESTRF?

PhracturedBlue:
Sorry to drop in like this, but I would probably do exactly that (put it in for dev only mode).  I was recently selecting a protocol on my Tx, and it actually takes a long time to scroll through and find the one you want these days.  I am actually thinking that if I can get back into working on Deviation, it probably makes sense to have a dedicated select-list page instead of the spinbox.

hexfet:
Nice to see your handle PB.  The PR is updated to only include TESTSER in dev builds.  Also added fix in UART_SetDataRate as discussed [here](https://www.deviationtx.com/forum/protocol-development/7474-serial-out-for-lrs-boxes?start=80#69082).

PR #400, opened 2018-05-05
More reliable and beep only once "beep" mux

vladislavy:


TheRealMoeder:
Do you see any problems in using the same code for mux voice?

vladislavy:
I can't test mux "voice". If you can test it, you can add the same code for voice with "voice_lock" latch.

PR #411, opened 2018-05-29
Add target emulator for t8sg t8sg_v2 t8sg_v2_plus

flagadajones:
Hi,
I've just add target emulator for : 
t8sg
t8sg-v2
t8sg-v2-plus


flagadajones:
juste for the screen color :-)  t8sg V2 plus have Yellow OLED screen and  t8sgV2 an LCD screen

TheRealMoeder:
For me it is not a good reason to keep it. If you prefer yellow, do the V2 emulator in yellow color. But maintaing another target just for the display color in the emulator is surely not worth it, IMHO.

vladislavy:
Anyway you can use devo7e-256 emulator with "hardware.ini" from respective transmitter (ir8m, t8sg, t8sg_v2, t8sg_v2_plus). The only difference will be incorrect switch mapping if you will import a model files from foreign transmitters. So I don't see great sense to have specific emulators for devo7e-256 clones.

flagadajones:
as you want but emulator is not a simulator 

an emulator definition extract 
> Emulators maintain the original look, feel, and behavior of the digital object, which is just as important as the digital data itself.  (Wikipedia)

PR #413, opened 2018-05-31
Branch 5.0

xtof-durr:
It should read
     M(x) = if(*Switch*) {  *Curve*( *Src* ) * *Scale* + *Offset*} else {0} + *Trim*
because Curve is a function, not a factor.


vladislavy:
You try merge branch 5.0 to master. Will be better to make changes at master branch.

PR #434, opened 2018-09-05
Protocol WFLY

goebish:
Add WFLY protocol, ported from MULTI
https://www.deviationtx.com/forum/protocol-development/7990-wfly-protocol


vladislavy:
Is something changed at "install_nightlies.tar.enc" ???

goebish:
I suppose no (why would the content of this file be modified unintentionally ?), but git didn't let me do a partial commit after I merged with master to fix conflicts in Makefile.inc
Do you prefer I make another PR ?

PR #444, opened 2018-09-28
fix nrf24l01 pin assigment in hardware.ini

nitram509:
While freshly installing the current 5.0.0 release,
I did get an error "missing modules nrf24l01".
This is because in default shipped hardware.ini,
the pin assignment in wrong.

This is a simple fix which hopefully solves a lot of people headaches.

Any feedback is welcome.

### DIFF
--- a/src/fs/common/hardware.ini
+++ b/src/fs/common/hardware.ini
@@ -23,6 +23,6 @@
 ; has_pa-a7105    = 1
 ; enable-cc2500   = A14
 ; has_pa-cc2500   = 1
-; enable-nrf24l01 = A14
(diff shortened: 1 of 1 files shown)


markusgritsch:
I don't think this solves any issues. It totally depends, on which pin you soldered the CS wire. For example, on my Devo 7E I soldered it to the TMS pin (because this was suggested by the build instructions I used), so I had to use A13.

nitram509:
hmm, ok, wasn't aware of this - I was under the assumption, there is a common sense (like best practice) what to solder where.
feel free to cancel or merge this PR.

Thanks for the feedback.

Issue #439, opened 2018-09-23
Mux "beep" with the voice mod.


Tygrys-1:
I have the voice mod installed on T8SG V2 plus.
I set the channel mixer to "beep" at a particular value.
When the value is reached on the channel then instead of the "beep" there is a "saving" word spoken.

When using the "voice" mux the correct sound file is used.

PR #450, opened 2018-10-12
Fix mux "beep" for extended audio

vladislavy:
It's related to issue #439

### DIFF
--- a/src/mixer.c
+++ b/src/mixer.c
@@ -350,7 +350,6 @@ void MIXER_ApplyMixer(struct Mixer *mixer, volatile s32 *raw, s32 *orig_value)
                 value = *orig_value - rate;
         }
         break;
-#if HAS_EXTENDED_AUDIO
     case MUX_BEEP:
         if (orig_value) {
             s32 new_value = raw[mixer->dest + NUM_INPUTS + 1];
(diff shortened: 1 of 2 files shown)


vladislavy:
We finally has not memory for Devo7e. This change doesn't fit to Devo7e memory.

TheRealMoeder:
That is the reason mux beep was only available in has extended audio

Tygrys-1:
> 
> 
> That is the reason mux beep was only available in has extended audio

It is not a problem to have both mux beep and voice. It is the beep to just do a "beep". As trimmers can do beep whith voice mod enabled I think that the mux beep can also do this.

TheRealMoeder:
This is rather small and we have room for it in the room now. Please rebase to recheck with travis ci

TheRealMoeder:
I just ran the Travis job again, so no need for a rebase. But with 100 bytes it is pretty pricey, to be honest.

Tygrys-1:
This make sense. It should be something doing a "beep".

TheRealMoeder:
MUSIC_KEY_PRESSING isn't defined in the default sound.ini, but MUSIC_MAXLEN has the same default beep in sound.ini as MUSIC_SAVING, so I think we should go for this. As vladislavy doesn't have the branch in his repository anymore, I will close this PR and open a new one.

PR #455, opened 2018-11-04
Protocol for Air Hogs "Star Trek" NCC-1701

goebish:
Ported from MULTI, untested on Deviation.

Protocol: NCC1701
Channel 5: Warp

I had to discard it from the modular build (7e) or it wouldn't fit in ROM (https://travis-ci.org/goebish/deviation/jobs/450493916), looks like adding only a new protocol definition overflows the allowed space by 20 bytes ... Note that it happens on Travis (arm-gcc 4.8.3 2013-q4) while I can still build it with the arm gcc version I'm using locally (arm-gcc 4.8.3 2014-q1). 
Perhaps we should change the version we're using on Travis (mine is only a few months newer) ? But the problem will still re-appear soon ...

Topic: https://www.deviationtx.com/forum/protocol-development/8101-star-trek-air-hogs


vladislavy:
I suggest to disable STANDARD GUI for Devo7e to free some memory.

goebish:
Let's wait until a decision is taken for STANDARD GUI on the 7e before merging this PR.

src/protocol/protocol.h
```diff
@@ -60,6 +60,9 @@ PROTODEF(PROTOCOL_DM002, NRF24L01, AETRG, DM002_Cmds, "DM002")
 PROTODEF(PROTOCOL_BUGS3MINI, NRF24L01, AETRG, BUGS3MINI_Cmds, "Bugs3Mini")
 PROTODEF(PROTOCOL_E012, NRF24L01, AETRG, E012_Cmds, "E012")
 PROTODEF(PROTOCOL_E015, NRF24L01, AETRG, E015_Cmds, "E015")
+#ifndef MODULAR
```
howard0su: remove this as devo7e has some bytes to use now

goebish:
Thanks @howard0su !
I've a few more protocols on hold, I'll make a pull request for them tomorrow.

src/protocol/ncc1701_nrf24l01.c
```diff
@@ -0,0 +1,402 @@
...
+    packet[9] = rx_id[2];
+    packet[10] = rx_id[3];
+    packet[11] = rx_id[4];
+    packet[12] = GET_FLAG(CHANNEL_WARP, 0x02);    // Warp:0x00 -> 0x02
+    packet[13] = packet[5]+packet[6]+packet[7]+packet[8]+packet[12];
+    if(phase == NCC_BIND_TX1)
+    {
+        packet[0] = 0xBB;
+        packet[5] = 0x01;
+        packet[6] = rx_tx_addr[2];
```
howard0su: can u reformat the code a little bit better? looks like when bind, you only need first 7 bytes. other protocol has the functions like build_bind_packet and build_data_packet.

goebish: it's ported from the MULTI project and I prefer keeping the code as it is, it will be easier to maintain.

TheRealMoeder:
Can you add it to the manual as well?

goebish:
I will

PR #462, opened 2018-11-20
Enable Modular Build for f12e

howard0su:
1. Fix modular build for frfkyx
2. Cleanup modular build and remove hack for uart.
3. Enable modular build for f12e via reserving 10K RAM


TheRealMoeder:
I have no f12e for testing, although code seems reasonable. Any particular reason to include wfly as a default protocol module?

victzh:
> I have no f12e for testing, although code seems reasonable. Any particular reason to include wfly as a default protocol module?

I agree, code seems reasonable.  I don't even know what wfly protocol is :-)

howard0su:
> > I have no f12e for testing, although code seems reasonable. Any particular reason to include wfly as a default protocol module?
> 
> I agree, code seems reasonable. I don't even know what wfly protocol is :-)

HAHA... it is because I have a F12E and only have three wfly receivers to play with. let's hold this change for a while because I am still struggle if this is right direction to enable modular build for f12e as it only have 64KB SPI.

I will split out some code change in this change and commit them first as they are anyway reasonable change.

PR #463, opened 2018-11-22
Enable LTO for non-modular build

howard0su:
LTO build will further optimize the code to save more ROM and RAM. Also fix some warnings noticed by LTO linker.

Use devof12e build as example:
Before LTO:
ROM: 0x08004000 - 0x08039650 = 213.58kB
RAM: 0x20000000 - 0x200068b8 =  26.18kB

After LTO:
ROM: 0x08004000 - 0x08037500 = 205.25kB
RAM: 0x20000000 - 0x200067c8 =  25.95kB

Difference:
ROM: 8.30kB
RAM: 0.25kB

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -159,8 +159,6 @@ ifeq "$(TYPE)" "dev"
 endif
 ifdef MODULAR
 CFLAGS   +=  -DMODULAR=$(MODULAR)
-else
-CFLAGS   += -flto
 endif
 
 CFLAGS   := $(CFLAGS) -Wall -Wextra -Werror=undef -I. -D$(PROGMODE) -std=gnu99 -Itarget/$(TARGET) -Igui/$(SCREENSIZE) \
--- a/src/Makefile
+++ b/src/Makefile
@@ -159,6 +159,8 @@ ifeq "$(TYPE)" "dev"
 endif
 ifdef MODULAR
 CFLAGS   +=  -DMODULAR=$(MODULAR)
+else
+CFLAGS   += -flto
 endif
 
 CFLAGS   := $(CFLAGS) -Wall -Wextra -Werror=undef -I. -D$(PROGMODE) -std=gnu99 -Itarget/$(TARGET) -Igui/$(SCREENSIZE) \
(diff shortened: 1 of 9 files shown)


howard0su:
yes. build is a bit slower but the ROM & RAM is smaller. I cannot figure out a way to expose the symbols for MODULAR build since LTO build will figure out PROTO_Stubs are not really running.

howard0su:
Fixed. I parse the elf file instead obj file for the final string list. I believe this should be faster and more accurate.

hexfet:
A build with the latest changes does not work for T8SG Plus.  The transmitter powers off as soon as the power button is released.  No sounds, no display.  A build from current master works fine.  I thought I tested a build from the previous commit prior to my last comment, but must've loaded the wrong file.  A build of 82f04018e for T8SG Plus does not work either.  The errors.txt file was empty which indicates the problem is not an unrecoverable exception.

vladislavy:
I suspect LTO trying to use variables with same names as single global variable. I seen warnings about use variables with different types, but same names. E.g. variable "state", u8 and u16, from different files. If variables with same name from different modules have the same type, we don't get any warning.
If this is the case, we can get many problems.

hexfet:
I tried with Devo10 with latest lto branch and the tx is bricked. Turn power switch on and nothing happens, and same with the EXT switch held down.  Guessing somehow the bootloader got corrupted.  Will track down how to recover.

Just to verify the test I tried the lto build on T8Sg Plus again, and same behavior as before.  Immediately powers off, but still able to access the bootloader.  Build from master works fine.

howard0su:
> I tried with Devo10 with latest lto branch and the tx is bricked. Turn power switch on and nothing happens, and same with the EXT switch held down. Guessing somehow the bootloader got corrupted. Will track down how to recover.
> 
> Just to verify the test I tried the lto build on T8Sg Plus again, and same behavior as before. Immediately powers off, but still able to access the bootloader. Build from master works fine.

sorry to hear that. I tried the bits on my f12e. I have to unplug battery and hold ent and turn on to get back update screen. I think boot loader should be fine.

I will close the PR until I figure out the problem. Don't want to more ppl to try this.

PR #464, opened 2018-11-24
Improve module build

howard0su:
Module Build is enhanced to make sure:
1. firmware code is built with the macro ENABLE_MODULAR.
2. The code used in module is built with macro MODULAR and force long jump. No code is mixed used in firmware or module.
3. Make sure other platform can enable modular to avoid hard code obj in ld command line like objs/devo7e/uart.o
4. Split out protospi code into a C file as the size of function is too large to inline in some compilers.


howard0su:
can I get review on this?

howard0su:
I have a f12e and I tested it with both modular and non-modular build with the change.

hexfet:
I think it's good, but don't have a limited memory tx.  Did you test SBUS or CRSF output with modular build?

What's the advantage of adding ENABLE_MODULAR?  When do I use it instead of MODULAR?

howard0su:
No. I am not familiar with such features. Let me research a bit.

ENABLE_MODULAR means the code is compile for main firmware but when MODULAR build is enabled (devo7e).
MODULEAR means the code is compile for module (of course MODULAR build is enabled).

I have to use two macros as some code need be compiled in both main firmware and module like uart.c.

hexfet:
Okay, I understand the reason.

However this change breaks the modular builds.  All the protocol files (*.mod) are zero length for both devo7e and devof7.  Building with the docker environment.

howard0su:
I revert the last change and rebase to master. Can u help test on Devo7e as I don't have that TX on hand.

hexfet:
Building 6bfb917b in the Docker environment still produces zero length protocol files.

hexfet:
The protocol files have reasonable sizes now.  Unfortunately I don't have a 7e or other tx to test modular build easily.

TheRealMoeder:
Did you check crsf/sbus serial output on trainer port?

howard0su:
no. How to do so?

howard0su:
i see. this is valid point. let me check.

howard0su:
I checked lst file for ISRs:
master build:
$ grep _isr devo7e.list
08005c44 <tim4_isr>:
08005cd0 <exti1_isr>:
0800602c <exti2_isr>:
080063dc <dma1_channel4_isr>:
08006418 <usart1_isr>:
080067a0 <tim1_up_isr>:
08006a3c <exti9_5_isr>:
08008210 <usb_lp_can_rx0_isr>:

this PR build:
$ grep _isr devo7e.list
08005c44 <tim4_isr>:
08005cd0 <exti1_isr>:
0800602c <exti2_isr>:
080063c8 <dma1_channel4_isr>:
08006404 <usart1_isr>:
0800678c <tim1_up_isr>:
08006a28 <exti9_5_isr>:
080081f4 <usb_lp_can_rx0_isr>:

for checking trainer output, I may need build a hardware. Is trainer port is TTL so that I can connect to USB-to-TTL adapter?

howard0su:
Okay. I will try to fix today.

hexfet:
You are right.  The CRSF protocol is broken for modular builds in master.  Testing with a modular build of T8+ with SBUS and TESTSER also enabled, the serial output did not work for any of them.

Then confirmed that moving the ISR code from uart.c to pwm_rom.c fixes the issue.  Judging by the name, this code gets located in ROM.

I've never fully dug into the modular file compile and link steps so need to do that to understand why the ISRs end up in the protocol modules.  There may not be any better solution than putting the ISRs in separate files.

Running as a module means the protocol code is not compiled into ROM, but into separate binaries with the .mod extension that are stored in the filesystem and loaded into RAM when the protocol is selected.  The PROTOCMD interface is the communication link to the dynamically loaded code.

howard0su:
This still doesn't explain well. I checked devo7.map, here is three ISRs related, (2 for uart, 1 for pwm). I also cannot find ISR code in crsf_uart.mod.

 2359  .text.dma1_channel4_isr
 2360                 0x08006b4c       0x3c objs/devo7e/uart.o
 2361                 0x08006b4c                dma1_channel4_isr
 2362  .text.usart1_isr
 2363                 0x08006b88       0x24 objs/devo7e/uart.o
 2364                 0x08006b88                usart1_isr
 2311  .text.tim1_up_isr
 2312                 0x0800662c       0x3c objs/devo7e/pwm_rom.o
 2313                 0x0800662c                tim1_up_isr

TheRealMoeder:
@hexfet so we can merge this as well?

howard0su:
need a bit time to cleanup protospi. 
static const struct mcu_pin CYRF_RESET_PIN ={GPIOB, GPIO11};
static const struct mcu_pin AVR_RESET_PIN ={GPIO_BANK_JTCK_SWCLK, GPIO_JTCK_SWCLK};

This part is really ugly to define a static in header file.

howard0su:
Done with cleanup

howard0su:
please squeeze when merge.

PR #465, opened 2018-11-24
fix frskyx_cc2500 in modular build when telemetry is enabled

howard0su:
modular build doesn't support data section. So we should not use initial value for static variables. Initialize it during UART setup.

### DIFF
--- a/src/protocol/frskyx_cc2500.c
+++ b/src/protocol/frskyx_cc2500.c
@@ -399,7 +399,10 @@ static u8 sport_crc(u8 *data) {
     return 0x00ff - crc;
 }
 
+static u8 outbuf[FRSKY_SPORT_PACKET_SIZE+2];
+
 static void setup_serial_port() {
+    outbuf[0] = 0x7e;
     if ( Model.proto_opts[PROTO_OPTS_SPORTOUT] ) {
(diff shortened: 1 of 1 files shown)


hexfet:
Your comment is correct, but this code is excluded from modular builds by the HAS_EXTENDED_TELEMETRY check at line 388.  I feel the intent is clearer with the code as it is.  None of the extended telemetry processing is included in modular builds.

howard0su:
this is noticed when I enable modular build for f12e. Any reason we cannot support extended telemetry in modular build?

goebish:
That's because the resulting .mod would be bigger than 4 kB.

howard0su:
> That's because the resulting .mod would be bigger than 4 kB.

I see. again devo7e only issue...

hexfet:
I'm not opposed to such a change (but I don't have an affected transmitter).

howard0su:
we don't have any TX so far enabled modular which also enable extended_telemetry. but this is code defect I noticed when I was trying to enable modular build for f12e. Later on, I want to hold on that change as f12e has really small SPI flash which may not be right direction to have modules.

hexfet:
If you have a future PR that requires this change as part of adding a feature or fixing an issue it can be merged then. Right now it does neither so my vote is no. (recognize I could be biased)

PR #466, opened 2018-11-24
Fix warnings under new gcc

howard0su:
Fix warnings notified by new gcc.
1. Access array out of range
2. explicit on fall through in switch branches.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -10,6 +10,7 @@
 #define ctassert(COND,MSG) typedef char static_assertion_##MSG[(COND)?1:-1]
 #define _UNUSED   __attribute__ ((unused))
 #define NO_INLINE __attribute__ ((noinline))
+#define FALLTHROUGH __attribute__ ((fallthrough))
 
 #define TEMPSTRINGLENGTH 400 //This is the max dialog size (80 characters * 5 lines)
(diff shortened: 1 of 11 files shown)


howard0su:
please help review the code. I intentionally didn't change any existing logic. But some state transition code in protocol is not super clear to me if it is right or not.

hexfet:
These changes cause new empty declaration warnings with the current compiler:
`protocol/kn_nrf24l01.c:278:9: warning: empty declaration [enabled by default]
         FALLTHROUGH;`

Not sure what you mean by "new gcc" but the version in the docker container is 4.8.3 20131129.  There has been some recent discussion about updating that version.

The [gcc manual page has some options](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wimplicit-fallthrough) for disabling the warning based on comments.  That might be a backwards-compatible way of addressing the warnings produced by later gcc versions.

tracernz:
You can also just have a comment, which has always good practice anyway.
```
case foo:
    do_thing(stuff);
    // fallthrough
case bar:
    unfudge(stuff);
    break;
```
GCC docs: https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wimplicit-fallthrough

PR #467, opened 2018-11-25
Remove standard GUI from Devo 7e build

goebish:
I've a few protocols on hold but the 7e won't accept any more protocol definitions, time to remove the standard GUI from the 7e build ?

### DIFF
--- a/src/target/devo7e/target_defs.h
+++ b/src/target/devo7e/target_defs.h
@@ -17,15 +17,15 @@
 #endif
 
 #define HAS_ADVANCED_GUI    1
-#define HAS_PERMANENT_TIMER 0
+#define HAS_PERMANENT_TIMER 1
 #define HAS_TELEMETRY       1
 #define HAS_EXTENDED_TELEMETRY 0
 #define HAS_TOUCH           0
 #define HAS_RTC             0
 #define HAS_VIBRATINGMOTOR  1
-#define HAS_DATALOG         0
+#define HAS_DATALOG         1
 #define HAS_SCANNER         0
-#define HAS_LAYOUT_EDITOR   0
+#define HAS_LAYOUT_EDITOR   1
 #define HAS_EXTRA_SWITCHES  OPTIONAL
 #define HAS_EXTRA_BUTTONS   0
 #define HAS_MULTIMOD_SUPPORT 1
--- a/src/target/devo7e/target_defs.h
+++ b/src/target/devo7e/target_defs.h
@@ -12,7 +12,8 @@
 //No room for debug and standard gui
  #define HAS_STANDARD_GUI   0
 #else
- #define HAS_STANDARD_GUI   1
+//Nor for production build 
+ #define HAS_STANDARD_GUI   0
 #endif
 
 #define HAS_ADVANCED_GUI    1


vladislavy:
We can enable some common features with it:

#define HAS_PERMANENT_TIMER 1
#define HAS_DATALOG         1
#define HAS_LAYOUT_EDITOR   1

TheRealMoeder:
I'm OK with this (though I'd rather add voice (without menu, due to ram limitations) than data log. I requested a review from pb, since he always refrained from removing standard gui. But if we don't hear anything from him within a week or so, I guess we might as well merge.

PhracturedBlue:
has anyone collected data from users about this?  are you sure that end users primarily use the standard gui?

goebish:
Seems most people have never heard of the standard GUI
https://www.deviationtx.com/forum/7-development/8108-remove-standard-gui-from-devo-7e-build

PhracturedBlue:
it may be interesting to check through the models saved in the wiki/forum and see how many use the standard GUI.  My bug concern is that users upgrade to something that is supposed to be newer and better, and in the process lose features.  That is never fun for anyone who gets hit.

I wish I had finished the binary config-file implementation in the uploader (compiling and decomiling config-files during upload rather than on the Tx).  Supporting the .ini files takes up a huge amount of space.  Another idea I had was to actually support dynamic libraries on the ROM (vs on the filesystem).  We could then allow users to choose which 'features' to load and ensure that they fit.  I don't know how much overhead such a system would take though.

However, my guess is that there is not enough developer interest to solve this problem for the potentially marginal gain.  I think my only request is that a last official release is made with current features before removing the standard GUI for the 7e so that users who like it have an official version they can use with the latest features.

howard0su:
I believe LTO may help reduce the ROM size. but I am still struggle to figure out why it doesn't work. 

We should limit the new features to put on devo7e even we can free up some ROM. Otherwise we will face the current situation very soon. We should face the reality that devo7e will not get all new/nice feature anymore due to ROM size limit.

PhracturedBlue:
i'll be very interested to see the results of LTO.  I also fully agree that devo7e shouldn't expect new features.  I am more concerned with removing existing ones at the moment though.  I would love a choose-your-build model, but it is very challenging to do without either precompiling lots of permutations or forcing everyone to be able to run a docker image (or maybe hosting a docker image that can take requests...)

TheRealMoeder:
Actually hosting a docker image  that can take requests is really simple, I did this for my deviation voicepacker [
https://moeder.pythonanywhere.com/voicepacker/default/index](url)

PhracturedBlue:
That is pretty cool.  If we could estimate the size of various features, we could probably use something like this to add a lot of flexibility for devo7e (and f7) builds

TheRealMoeder:
Well, at first it was merely a proof of concept to myself, and the free pythonanywhere account is somewhat limited, but simple as it is it works quite well. Simple api calls to travis from the Python page with parameters passed, and the results being pushed to githubio. Running python on deviationtx.com would make room for more stuff. Most useful probably saving the build and providing a hashed Download link...

howard0su:
I finally get 7e build under LTO. here is the size:
ROM: 0x08003000 - 0x0801eee4 = 111.72kB
RAM: 0x20000000 - 0x200027a4 =   9.91kB

Compare to non-LTO:
ROM: 0x08003000 - 0x08020000 = 116.00kB
RAM: 0x20000000 - 0x20002968 =  10.35kB

But I still need figure out why initialize fails in LTO build. I need some help here if anyone has SWD/JTAG attached to TX. I need additional couple of days to get my device to install a SWD pins.

howard0su:
i was wrong. LTO for devo7e get overflow in ROM still.

TheRealMoeder:
Considering @howard0su 's work, what do you think of branching vlads work as version 5.1 and remove standard gui for nightly and future 6.0?

hexfet:
Sounds like a possible way forward.  Does vlad's branch include non-protocol changes to master?

Will anyone miss the 5.0 GUI?

PR #469, opened 2018-11-27
Refactor Font code to be in a seperate module

howard0su:
Create a separate module FONT to handle .fon file inside SPI flash.


TheRealMoeder:
And yet again we hit the devo7e flash ROM limit...

PhracturedBlue:
what is the reason for this change?  the fonts really are specific to the tx model (or specifically screen type).  the f7 for instance has hardcoded fonts and doesn't use the dynamic fonts

TheRealMoeder:
> what is the reason for this change? the fonts really are specific to the tx model (or specifically screen type). the f7 for instance has hardcoded fonts and doesn't use the dynamic fonts

Probably this: [
https://www.deviationtx.com/forum/7-development/8135-mini-font-for-cn-tw-ru](url)

PhracturedBlue:
I don't see why this complexity is needed on all platforms.  I don't mind the solution, but think it should be isolated to the F series Tx (and not moved to the gui or screen dirs).  if there is actually common code, it can be accessed from the common dir.  A solution tlike that shouldn't impact the dveo7e

howard0su:
Font code is handle .fon file which should be a component shared across Tx. It is not specific for F serial. The new code is clean. Look at font.h, the API is pretty isolated.

Font code today is not clear isolated with LCD driver. but in theory, they should be different. I would like to clean up them for future reuse.

The size change should related to some function is not inlined anymore. It should be fixed after I figure out LTO build.

howard0su:
close this for now. Seems I don't need a common font implementation for f12e localization support.

howard0su:
This may needed for devo7e for localization support. keep it for now

PR #471, opened 2018-11-29
Add Linux emulation to travis dailybuild

howard0su:


howard0su:
I don't use Windows to debug. So I am using Linux version of emulator. I think Linux version of Emulator is useful. And we should not let it get broken.

PR #472, opened 2018-12-01
Add PXX protocol

hexfet:
The PXX protocol is an Frsky serial interface to their transmitter modules.  Primarily useful for R9M and XJT modules in the JR bay of the T8SG Plus, but protocol is available on the trainer jack of other transmitters.  Telemetry (s.port) is supported.

Protocol options available to specify failsafe (see FrskyX manual entry for details), country (must match radio module firmware), and receiver telemetry and pwm output controls.  Range check operation is supported.  Module power level settings correspond to top four deviation power settings.

Due to memory constraints PXX is available only in non-modular builds.

This PR also includes changing the PPM implementation to use DMA instead of interrupts.  Reduces overall interrupt load, and was low-hanging fruit because DMA is required for PXX.


src/target/common/devo/protocol/pwm.c
```diff
@@ -79,29 +69,120 @@ void PWM_Stop()
...
+
+    u8 pxx_bit = 1 << 7;
+    u8 pxx_ones_count = 0;
+    u8 pxx_bytes = 18;
+    u8 *pc = pxx_bits;
+
+    // flag bytes are not bit-stuffed
+    memcpy(pc, pxx_flag, 8);
+    pc += 8;
+    for (;;) {
```
howard0su: use for loop here. it is recommended never use infinite loop in firmware to avoid possible hang. the logic here can be:
for(pxx_bytes=18;pxx_bytes>0;pxx_bytes = pxx_bytes-1)

src/target/common/devo/protocol/pwm.c
```diff
@@ -18,58 +18,48 @@
...
 #endif
 
-#include "common.h"
-#ifndef DISABLE_PWM
 
 #include <libopencm3/stm32/rcc.h>
-#include <libopencm3/stm32/gpio.h>
 #include <libopencm3/stm32/timer.h>
+#include <libopencm3/stm32/gpio.h>
+#include <libopencm3/stm32/dma.h>
```
howard0su: sorry includes please.

src/protocol/pxxout.c
```diff
@@ -0,0 +1,372 @@
...
+               | (Model.proto_opts[PROTO_OPTS_RXPWM] << 2);
+
+    u16 lcrc = crc(packet, PXX_PKT_BYTES-2);
+    packet[16] = lcrc;
+    packet[17] = lcrc >> 8;
+}
+
+static enum {
+  PXX_BIND,
+#ifndef EMULATOR
```
howard0su: why we need this?

src/protocol/pxxout.c
```diff
@@ -0,0 +1,372 @@
...
+  0x4a44,0x5bcd,0x6956,0x78df,0x0c60,0x1de9,0x2f72,0x3efb,
+  0xd68d,0xc704,0xf59f,0xe416,0x90a9,0x8120,0xb3bb,0xa232,
+  0x5ac5,0x4b4c,0x79d7,0x685e,0x1ce1,0x0d68,0x3ff3,0x2e7a,
+  0xe70e,0xf687,0xc41c,0xd595,0xa12a,0xb0a3,0x8238,0x93b1,
+  0x6b46,0x7acf,0x4854,0x59dd,0x2d62,0x3ceb,0x0e70,0x1ff9,
+  0xf78f,0xe606,0xd49d,0xc514,0xb1ab,0xa022,0x92b9,0x8330,
+  0x7bc7,0x6a4e,0x58d5,0x495c,0x3de3,0x2c6a,0x1ef1,0x0f78
+};
+
+static u16 crc(u8 *data, u8 len) {
```
howard0su: we should consider to move crc16 code out. there are multi implementation of crc16 in the code already.

src/protocol/pxxout.c
```diff
@@ -0,0 +1,372 @@
...
+};
+ctassert(LAST_PROTO_OPT <= NUM_PROTO_OPTS, too_many_protocol_opts);
+
+#define FAILSAFE_HOLD    0
+#define FAILSAFE_NOPULSE 1
+#define FAILSAFE_RX      2
+
+
+static u8 packet[PXX_PKT_BYTES];
+u8 num_channels;
```
howard0su: why this is not static (local variable)?

src/target/common/devo/protocol/pwm.c
howard0su: typo in my comment

hexfet: Seems like a good idea, though I couldn't find much about best practices online.  Reordered in the spirit of [these guidelines.](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes)

src/target/common/devo/protocol/pwm.c
```diff
@@ -79,29 +68,117 @@ void PWM_Stop()
...
+    timer_generate_event(TIM1, TIM_EGR_UG);      // Generate update event to start DMA
+    timer_enable_counter(TIM1);
+}
+
+#ifndef MODULAR
+#define PXX_PKT_BYTES 18
+#define PXX_PKT_BITS  (20 * 8 + 32)  // every 5th bit might be escaped
+static const u8 pxx_flag[] = {15, 23, 23, 23, 23, 23, 23, 15};  // 7e
+static u8 pxx_bits[PXX_PKT_BITS];
+void PXX_Enable(u8 *packet)
```
howard0su: why this function is defined here instead of in the same file as caller?

hexfet: Follows the example of PPM_Enable.  Functions with dependencies on the libopencm3 library should be in the target/ tree.

Considered putting the bytes-to-bits conversion in the protocol file but it fits better here with the dma initialization.

howard0su: i see .

src/protocol/protocol.c
```diff
@@ -405,7 +405,11 @@ int PROTOCOL_HasPowerAmp(int idx)
     } else {
 #endif        
     int m = get_module(idx);
-    if(m != TX_MODULE_LAST && Transmitter.module_poweramp & (1 << m))
+    if ((m != TX_MODULE_LAST && Transmitter.module_poweramp & (1 << m))
+#ifndef MODULAR
+        || (Model.protocol == PROTOCOL_PXX)
```
howard0su: why this is model is so special here?

src/protocol/testser.c
```diff
@@ -64,9 +64,10 @@ void check_options(u8 init) {
         speed = Model.proto_opts[TESTSER_SPEED];
         switch(speed) {
         case 0: UART_SetDataRate(115200); break;
-        case 1: UART_SetDataRate(9600); break;
-        case 2: UART_SetDataRate(100000); break;
-        case 3: UART_SetDataRate(400000); break;
+        case 1: UART_SetDataRate(57600); break;
```
howard0su: can we keep the original order but add 57600 as last one? It is always good to not change user experience unless we really need.

src/protocol/protocol.h
```diff
@@ -68,6 +68,7 @@ PROTODEF(PROTOCOL_USBHID, TX_MODULE_LAST,  NULL, USBHID_Cmds, "USBHID")
 PROTODEF(PROTOCOL_CRSF, TX_MODULE_LAST,  NULL, CRSF_Cmds, "CRSF")
 #ifndef MODULAR
 PROTODEF(PROTOCOL_SBUS, TX_MODULE_LAST,  NULL, SBUS_Cmds, "SBUS")
+PROTODEF(PROTOCOL_PXX, TX_MODULE_LAST,  NULL, PXXOUT_Cmds, "PXX")
```
howard0su: we can support this define in devo7e as there are bytes to use.

TheRealMoeder: I this thought sbus and pxx modules are larger than 4kb

src/protocol/testser.c
TheRealMoeder: This is only used for developer purposes anyways. We might as welk sort ascending or descending by bps

src/protocol/protocol.c
hexfet: It's because we don't have the uart defined as a radio module, so it can't have a poweramp setting.  Adding code for a uart radio module didn't seem like good solution because the actual radios installed in the JR module bay can be different.  Checking for PXX protocol is a special case solution.  If we add support for other uart radio modules it will probably be best to look for a more general solution.

src/protocol/testser.c
hexfet: And TESTSER doesn't work in the dev build because dev grabs the serial port for printf.  There's really not much need for TESTSER since SBUS and CRSF are now available.  The only advantage of TESTSER is the loopback mode.

src/protocol/pxxout.c
howard0su: this table is exact same as the one in protocol/frskyx_cc2500.c

src/target/devo12/devo12.ld
```diff
@@ -1,6 +1,6 @@
 MEMORY
 {
-	rom (rx) : ORIGIN = 0x08004000, LENGTH = 240K  /* 16K is allocated to the bootloader, so only 240K is available */
+	rom (rx) : ORIGIN = 0x08004000, LENGTH = 496K  /* 16K is allocated to the bootloader, so only 496K is available */
 	ram (rwx) : ORIGIN = 0x20000000, LENGTH = 48K
```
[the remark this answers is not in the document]
TheRealMoeder: You should probably post a build in the forum, there should be people willing to test the rom usage increase

src/protocol/protocol.h
hexfet: The pxx module is less than 1.5K, and sbus is less because it lacks telemetry.  The issue with pxx is the soft serial support.  It uses interrupts so whatever is the issue with the uart code is likely to affect pxx telemetry too.

src/target/devo12/devo12.ld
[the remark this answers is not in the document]
hexfet: Made a new test build and posted about it.

vladislavy:
If test build is [hexfet]protocol_pxx deviation-devo12-v5.0.0-bb6c3ca, it can be flashed with Dfuse and it works.

PR #473, opened 2018-12-02
Move channel map into individual module

howard0su:
ChannelMap is always reference with Model.Protocol. Replace it with a CurrentChannelMap and leave the logic to decide the channel map to individual module. This save 244bytes from devo7e.
ROM: 0x08003000 - 0x0801ff10 = 115.77kB
RAM: 0x20000000 - 0x20002894 =  10.14kB

Didn't remove map from protocol.h where it is not used anymore. Use it as a documentation.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -129,7 +129,7 @@ enum Protocols {
     PROTOCOL_COUNT,
 };
 #undef PROTODEF
-extern const u8 *ProtocolChannelMap[PROTOCOL_COUNT];
+extern const u8 *CurrentProtocolChannelMap;
 #define PROTO_MAP_LEN 5
 
 enum ModelType {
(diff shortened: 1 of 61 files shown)


howard0su:
Please help testing the change as I don't have devo7e.

howard0su:
good catch. I fixed in latest change.

TheRealMoeder:
I would suggest adding PROTOCMD-CHANNELDEF to all protocols to avoid confusion

howard0su:
I was thinking about that. but I didn't do that rather leaving a default value. Because it will add code size when we do a non-modular build. what's your suggetion?

TheRealMoeder:
How about adding it commented out with the advice, that this is done behause it is the default mapping?

howard0su:
sounds good.

howard0su:
I prefer squeeze the changes into one when merge this PR.

howard0su:
Didn't pass the testing on the real hardware devo7e. Need debug.

howard0su:
Rebase with the master and fix the issue on devo7e.

I think this is ready to go. Please squeeze the changes when checkin.

howard0su:
need code review. thanks.

howard0su:
@hexfet please review.

howard0su:
rebased with master and cleanup some whitespace change.

PR #474, opened 2018-12-02
rework the way to expose protocol data

howard0su:
Use an array Protocols to expose various data to avoid big switch in the code.

This saves another 492bytes from devo7e build.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -124,13 +124,12 @@ int PAGE_ModelDoneEditing();
 /* Protocol */
 #define PROTODEF(proto, module, map, init, name) proto,
 enum Protocols {
-    PROTOCOL_NONE,
+    PROTOCOL_NONE = 0,
     #include "protocol/protocol.h"
     PROTOCOL_COUNT,
 };
 #undef PROTODEF
 extern const u8 *ProtocolChannelMap[PROTOCOL_COUNT];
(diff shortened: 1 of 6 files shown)


howard0su:
Fixed in the latest change.

howard0su:
I plan to buy a devo7e so that I can test by myself.

howard0su:
anyway, Just bought a second hand dev7e which only costs me $25.

howard0su:
devo7e cannot do binding. need fix. please hold.

howard0su:
Verified on devof12e and devo7e. I think this is ready to go.

howard0su:
@hexfet can you enable modular build for t8sg with HAS_EXTENDED_TELEMETRY=1? I think you should be able to do so after apply PR #465.

hexfet:
I made a modular build for T8+ since my T8 is in pieces.   Here's the [changeset](https://github.com/DeviationTX/deviation/compare/master...hexfet:t8sg_plus_modular).  The build is limited to 128K ROM to match builds for limited memory transmitters.

Even with full ROM size enabling extended telemetry for the modular build fails because then some protocols won't fit into 4K of RAM.  DSM and FrskyX are two examples.

howard0su:
i see. i worked a change for f12e to enable 8K RAM for modules. anyway, i think I understand your intention here.

PR #475, opened 2018-12-05
Make PROTO_TELEM_UNSUPPORTED to 0

howard0su:
As most protocol doesn't support telemetry, make unsupport as 0 to save
ROM size 264 bytes. Also fixed the places which assumes the values.

For example of devof12e:
before:
ROM: 0x08004000 - 0x0803c13c = 224.31kB
RAM: 0x20000000 - 0x200069f8 =  26.49kB

After:
ROM: 0x08004000 - 0x0803c034 = 224.05kB
RAM: 0x20000000 - 0x200069f8 =  26.49kB

### DIFF
--- a/src/telemetry.c
+++ b/src/telemetry.c
@@ -330,7 +330,7 @@ void TELEMETRY_SetType(int type)
 //#define DEBUG_TELEMALARM
 void TELEMETRY_Alarm()
 {
-    if (! PROTOCOL_GetTelemetryState())
+    if (PROTOCOL_GetTelemetryState() == PROTO_TELEM_UNSUPPORTED)
         return;
 
     //Update 'updated' state every time we get here
--- a/src/telemetry.c
+++ b/src/telemetry.c
@@ -330,7 +330,7 @@ void TELEMETRY_SetType(int type)
 //#define DEBUG_TELEMALARM
 void TELEMETRY_Alarm()
 {
-    if (PROTOCOL_GetTelemetryState() == PROTO_TELEM_UNSUPPORTED)
+    if (PROTOCOL_GetTelemetryState() != PROTO_TELEM_ON)
         return;
 
     //Update 'updated' state every time we get here
--- a/src/telemetry.h
+++ b/src/telemetry.h
@@ -218,9 +218,9 @@ struct Telemetry {
 };
 
 enum {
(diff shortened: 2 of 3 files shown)


src/telemetry.c
```diff
@@ -330,7 +330,7 @@ void TELEMETRY_SetType(int type)
 //#define DEBUG_TELEMALARM
 void TELEMETRY_Alarm()
 {
-    if (! PROTOCOL_GetTelemetryState())
+    if (PROTOCOL_GetTelemetryState() == PROTO_TELEM_UNSUPPORTED)
```
PhracturedBlue: I don't think this is right   shouldn't it be == PROTO_TELEM_OFF

src/telemetry.h
```diff
@@ -218,9 +218,9 @@ struct Telemetry {
 };
 
 enum {
-    PROTO_TELEM_UNSUPPORTED = -1,
-    PROTO_TELEM_OFF = 0,
-    PROTO_TELEM_ON  = 1,
+    PROTO_TELEM_UNSUPPORTED = 0,
+    PROTO_TELEM_OFF = 1,
+    PROTO_TELEM_ON  = 2,
 };
```
PhracturedBlue: I haven't looked at it in detail, but will this break existing model files?

TheRealMoeder: It doesn't seem to be model file related. It is only used to activate telemetry alarms and in pages

src/telemetry.c
TheRealMoeder: Shouldn't it be `if (!(PROTOCOL_GetTelemetryState() == PROTO_TELEM_ON))`?

hexfet: Correct.  Only have telemetry values to alarm on when state is ON.

src/telemetry.h
hexfet: Shouldn't break models.  Looks like all the protocols use local defines for the protocol option values.

src/telemetry.c
howard0su: You are right. I am still wondering how can I better test this without real hardware?

TheRealMoeder: @hexfet I‘m sorry, it was late for me ;)

hexfet: No worries, everybody does it :)

PR #481, opened 2018-12-11
Don't use image size to create mapped gfx

howard0su:
first, the code cannot handle any size other than 6x4.
second, the unit of image size is pixel, while mapped gfx window is
char. They cannot use mixed. Let's always use 6x4 for devof12e.

### DIFF
--- a/src/pages/common/_main_page.c
+++ b/src/pages/common/_main_page.c
@@ -293,11 +293,13 @@ void GetElementSize(unsigned type, u16 *w, u16 *h)
         [ELEM_BATTERY]  = BATTERY_H,
         [ELEM_TXPOWER]  = TXPOWER_H,
     };
+#ifndef _DEVO_F12E_TARGET_H_
     if (type == ELEM_MODELICO && Model.icon[0]) {
         if(LCD_ImageDimensions(CONFIG_GetCurrentIcon(), w, h))
             return;
         //We can't fix this during model-load because only 1 file can be open at a time
     }
+#endif
     *w = width[type];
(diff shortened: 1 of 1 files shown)


TheRealMoeder:
Without reading the code, doesn't this apply to all text gui transmitters?

howard0su:
No. Only devof12e

howard0su:
Looks like this only applies to f12e, as only f12e implemented mapped_gfx.
docker@6a6e5a8a95f4:/git/deviation/src$ git grep HAS_MAPPED_GFX
gui/text/_gui.c:#if HAS_MAPPED_GFX
gui/text/_mapped_gfx.h:#if ! HAS_MAPPED_GFX
gui/text/_mapped_gfx.h:#endif //HAS_MAPPED_GFX
pages/text/toggle_select.c:#if HAS_MAPPED_GFX
screen/text/lcd_gfx.c:#if HAS_MAPPED_GFX
screen/text/lcd_gfx.c:  #if HAS_MAPPED_GFX
target/devof12e-XMS/target_defs.h:#define HAS_MAPPED_GFX      1
target/devof12e/target_defs.h:#define HAS_MAPPED_GFX      1
target/devof4-XMS/target_defs.h:#define HAS_MAPPED_GFX      0
target/devof4/target_defs.h:#define HAS_MAPPED_GFX      0
target/devof7-XMS/target_defs.h:#define HAS_MAPPED_GFX      0
target/devof7/target_defs.h:#define HAS_MAPPED_GFX      0

PR #482, opened 2018-12-11
New LCD emulation for devof12e

howard0su:
Use a low level emulatioin for LCD controller in devof12e so that we can
test LCD.c code in emulator.

While I am here, cleanup LCD.c to only contains the high level logic and leave the logic interfacing TW8816 chip to tw8816.c


howard0su:
This is ongoing work to support localization for devof12e.

howard0su:
tested on devof12e and emu_devof12e.

TheRealMoeder:
This seems to have broken nightlies...

src/target/devof12e/tw8816.c
```diff
@@ -29,19 +31,230 @@
...
+}
+
+static u32 LCD_ReadReg(unsigned reg)
+{
+    u8 val;
+    for (int i = 0; i < 2; i++) {
+        if (! I2C1_ReadBuffer(0x45, &val, reg, 1))
+            break;
+    }
+    return val;
```
TheRealMoeder: We still get a compiler warning here about `val` possibly being uninitialized

howard0su: i will fix it in #503

howard0su: already fixed in master.

PR #487, opened 2018-12-16
Uart modular fix

hexfet:
Fix the UART based protocols in modular builds.  Splits the uart.c code into three files to produce a modular build that works correctly for CRSF and SBUS.  Would like a better solution than three separate files but haven't yet figured out a better solution.  Making this PR so others can see and hopefully spot why the uart doesn't transmit unless the code is organized this way.

SBUS added to modular build by this PR.


hexfet:
This fix works but I can't explain why.  The UART_Send function is being located by the linker to the protocol .mod file.  It gets a valid RAM address, just like PWM_Initialize, so that doesn't seem to be a problem.  But if the ISRs are not put into a separate file then serial output doesn't work. 

The UART_Send function is split into the target/common/devo/protocol directory because the makefile builds files in that directory for both modular and non-modular builds, but in the modular build those files are built during the protocol compilations.  The linker puts the UART_Send function in RAM instead of locating it in ROM and referencing it from the .mod code.  That's what's done for the other UART_ functions, but again not sure why.

Following is from a modular build of T8+.
```
/src$ readelf -s --dyn-syms t8sg_v2_plus.elf|grep UART_
  3108: 08006b45    32 FUNC    GLOBAL DEFAULT    1 UART_SetDataRate
  3165: 08006bd9    48 FUNC    GLOBAL DEFAULT    1 UART_StartReceive
  3216: 08006c09     6 FUNC    GLOBAL DEFAULT    1 UART_StopReceive
  3224: 08006c6d   248 FUNC    GLOBAL DEFAULT    1 UART_Initialize
  3251: 08006c55    24 FUNC    GLOBAL DEFAULT    1 UART_SetDuplex
  3477: 08006b65   116 FUNC    GLOBAL DEFAULT    1 UART_SetFormat
  3835: 08006c11    68 FUNC    GLOBAL DEFAULT    1 UART_Stop
/src$ readelf -s --dyn-syms objs/t8sg_v2_plus/crsf_uart.elf|grep UART_
   101: 08006bd9     0 NOTYPE  LOCAL  DEFAULT  ABS UART_StartReceive
   144: 08006c09     0 NOTYPE  LOCAL  DEFAULT  ABS UART_StopReceive
   348: 08006b65     0 NOTYPE  LOCAL  DEFAULT  ABS UART_SetFormat
   641: 08006c11     0 NOTYPE  LOCAL  DEFAULT  ABS UART_Stop
   898: 08006b45     0 NOTYPE  GLOBAL DEFAULT  ABS UART_SetDataRate
   910: 08006c6d     0 NOTYPE  GLOBAL DEFAULT  ABS UART_Initialize
   913: 08006c55     0 NOTYPE  GLOBAL DEFAULT  ABS UART_SetDuplex
   948: 2000420d   148 FUNC    GLOBAL DEFAULT    1 UART_Send
/src$ readelf -s --dyn-syms objs/t8sg_v2_plus/ppmout.elf|grep PWM_
   900: 20004151   240 FUNC    GLOBAL DEFAULT    1 PWM_Initialize
   958: 20004241    80 FUNC    GLOBAL DEFAULT    1 PWM_Stop
```

src/target/common/devo/uart_rom.c
```diff
@@ -0,0 +1,49 @@
...
+    usart_disable_tx_dma(_USART);
+    dma_disable_channel(_USART_DMA, _USART_DMA_CHANNEL);
+
+    busy = 0;
+}
+
+usart_callback_t *rx_callback;
+void _USART_ISR(void)
+{
+	u8 status = USART_SR(_USART) & (USART_SR_RXNE | USART_SR_PE | USART_SR_FE | USART_SR_NE | USART_SR_ORE) ;
```
howard0su: whitespace

src/target/common/devo/protocol/uart_ram.c
```diff
@@ -0,0 +1,43 @@
+/*
+    This project is free software: you can redistribute it and/or modify
```
howard0su: where this file get referenced? when we compile mod, we don't auto include this file.

src/target/common/devo/Makefile.inc
```diff
@@ -78,7 +78,7 @@ $(TARGET).dfu: $(TARGET).bin
 ###################################
 $(ODIR)/%.bin: $(ODIR)/%.o_ $(PROTO_LINKFILE) $(PROTO_EXTRA_OBJS) $(TARGET).bin
 	@echo " + Building '$@'"
-	$(LD) -o $(@:.bin=.elf) $< $(PROTO_EXTRA_OBJS) objs/devo7e/uart.o $(LIBOPENCM3) -gc-sections -Map=$(@:.bin=.map) \
+	$(LD) -o $(@:.bin=.elf) $< $(PROTO_EXTRA_OBJS) $(LIBOPENCM3) -gc-sections -Map=$(@:.bin=.map) \
```
howard0su: I suggest first merge #464 as this is also changed in 464 in a more cleaner way.

src/target/common/devo/protocol/uart_ram.c
hexfet: Files in this directory are included in the .mod builds for modular builds, and also in a normal build.  See lines 24-32 of target/common/devo/Makefile.inc.

howard0su: okay. I will change my PR #464

src/target/common/devo/uart_rom.c
hexfet: Fixed whitespace issues.

TheRealMoeder:
Isn't this a similar issue with the interrupt vector you fixed for PWM with commit 83ebccf?

howard0su:
I believe we can only figure out the root cause is attach a JTAG to one TX and debug from there. Maybe I will find some time to do so in holiday.

howard0su:
@hexfet the change on modular build is checked in. Now the ISR code can be wrapper under the macro to make sure it is only compiled for main firmware when it is a modular build. So you may not need split into two files now.

TheRealMoeder:
Shall we merge this soon to fix uart?

howard0su:
Part of LTO work, I split out all ISR related code to a separate file which
is similar with this change. Let me test that change and if that one fix
the UART as well, I would prefer just merge LTO change without really
enable LTO. Since anyway, LTO work need that change.

hexfet:
Issue was fixed by #599.

PR #491, opened 2018-12-18
Fix mux beep for "extended_audio"

TheRealMoeder:
Fix for issue #439 
See also the discussion in closed PR #450


src/main.c
```diff
@@ -211,7 +211,7 @@ void medium_priority_cb()
 void EventLoop()
 {
     CLOCK_ResetWatchdog();
-#if !HAS_EXTENDED_AUDIO
+#if !HAS_EXTENDED_AUDIO && (!defined(HAS_HARD_POWER_OFF) || !HAS_HARD_POWER_OFF)
     unsigned int time;
```
howard0su: move time to the place you really use line 245. this simplied #ifdef logic here.

Issue #495, opened 2018-12-19
[TODO] Cleanup Music API usage in protocols


howard0su:
There are couple of protocols are using MUSIC APIs. this is not right behavior as protocol should not impact the user experience. Someone need analyze the needs for protocols and expose some APIs from PROTOCOL apis for those protocols to consume.

I believe most of MUSIC API calls are not needed as most of them are play music of Done_Binding which should be played by the logic inside protocol.c.

$git grep MUSIC protocol
protocol/bayang_nrf24l01.c:        // MUSIC_Play(MUSIC_TELEMALARM1);    // Shouldn't play telemetry alarm doing bind init
protocol/bayang_nrf24l01.c:            MUSIC_Play(MUSIC_DONE_BINDING);
protocol/bugs3_a7105.c:        MUSIC_Play(MUSIC_ALARM1);
protocol/cflie_nrf24l01.c:            MUSIC_Play(MUSIC_DONE_BINDING);
protocol/cg023_nrf24l01.c:        // MUSIC_Play(MUSIC_TELEMALARM1);     // Shouldn't play telemetry alarm doing bind init
protocol/cg023_nrf24l01.c:            MUSIC_Play(MUSIC_DONE_BINDING);
protocol/cx10_nrf24l01.c:        // MUSIC_Play(MUSIC_TELEMALARM1);      // Shouldn't play telemetry alarm doing bind init
protocol/cx10_nrf24l01.c:            MUSIC_Play(MUSIC_DONE_BINDING);
protocol/esky150_nrf24l01.c:            MUSIC_Play(MUSIC_DONE_BINDING);
protocol/esky_nrf24l01.c:        // MUSIC_Play(MUSIC_TELEMALARM1);      // Shouldn't play telemetry alarm doing bind init
protocol/esky_nrf24l01.c:            MUSIC_Play(MUSIC_DONE_BINDING);
protocol/fy326_nrf24l01.c:        // MUSIC_Play(MUSIC_TELEMALARM1);     // Shouldn't play telemetry alarm doing bind init
protocol/fy326_nrf24l01.c:            MUSIC_Play(MUSIC_DONE_BINDING);
protocol/fy326_nrf24l01.c:        // MUSIC_Play(MUSIC_TELEMALARM1);     // Shouldn't play telemetry alarm doing bind init
protocol/fy326_nrf24l01.c:            MUSIC_Play(MUSIC_DONE_BINDING);
[...]

PR #497, opened 2018-12-19
Remove MUSIC_Play from protocol

howard0su:
Fix issue #495


src/protocol/protocol.c
```diff
@@ -402,7 +402,6 @@ void PROTOCOL_CheckDialogs()
             proto_state |= PROTO_BINDDLG;
         } else if (proto_state & PROTO_BINDDLG) {
             PAGE_CloseBindingDialog();
-            MUSIC_Play(MUSIC_DONE_BINDING);
```
goebish: This one should be kept.

howard0su: ok

PR #503, opened 2018-12-20
Cleanup variable declaration

howard0su:
Make sure we only declare variables in header files. Move allocation to .c file. Also declare the local variables with "static".

### DIFF
--- a/src/config/model.h
+++ b/src/config/model.h
@@ -17,10 +17,10 @@
 #endif
 
 /* INI file consts */
-const char *MODEL_NAME;
-const char *MODEL_ICON;
-const char *MODEL_TYPE;
-const char *MODEL_TEMPLATE;
+extern const char *MODEL_NAME;
+extern const char *MODEL_ICON;
+extern const char *MODEL_TYPE;
+extern const char *MODEL_TEMPLATE;
 #define UNKNOWN_ICON ("media/noicon" IMG_EXT)
 
 //This cannot be computed, and must be manually updated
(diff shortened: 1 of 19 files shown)


howard0su:
@goebish can u help testing the change 9c0c37c as I don't have hardware to do so. This trick can help reduce the ROM size about 50bytes. I can see such practice in many protocols. We need consider to fix it.

goebish:
I tested with a Blade Inductrix, it works but default channel order is incorrect (probably not because of this PR), it should be TAER for DSM2/X but it's AETR now:
https://imgur.com/a/3pUC3gO

howard0su:
I can be caused by my other change on channel mapping. let me check.

goebish:
After resetting the model it's defaulting to EATR now, which is still wrong.

howard0su:
cannot find obvious error. both dsm code and protocol.c code looks correct. need debug further.

goebish:
I confirm that a version dating from 1 week ago have correct channel mapping for DSM2/DSMX.

howard0su:
I think i know why. Original the channel map doesn't require load the module. After my change, the channel map have to be loaded from module. In module configuration menu, we select DSMX then exit from the menu. The module is *not* loaded at all. So the channel map is incorrect. If you do BIND in the menu, the channel mapping is correct then.

We cannot delay load the module. Let me think about how to fix this.

howard0su:
check #504 which should fix channel map problem.

PR #504, opened 2018-12-20
Set channel mapping during PROTOCOL_Load

howard0su:
We need channel mapping after we load the protocol, and before initialize the protocol. Fix this to align with the original logic.


howard0su:
this is a regression from my change on channel mapping. Please review and merge.

PR #507, opened 2018-12-21
Support color in DevoF12E

howard0su:
1. Use invert color to display selected item.
2. update emu as well.


howard0su:
needs more testing in real Tx.

PR #508, opened 2018-12-22
Fix fltk build after migrate to latest osx

howard0su:
Latest OSX mojava has the build failure. Fix it via passing the correct CFLAGS and LDFLAGS.


howard0su:
can i get this in? This is pretty straightforward change.

PR #511, opened 2018-12-23
New Format of localization File

howard0su:
This change introduced a new file format for lang.* under language folder. The file is formatted in the way:
Line 1: <BOM>LanguageName
Line2-N: <16bit FNV Hash of english string>Localized String

The change increases the performance switching the language as we don't only read half of the content and don't need calculate hash as the hash is compute during build.
The change significantly reduces the size of localized file. After this change, I can successfully pack all localization files into 64KB SPI flash on F12E.

### DIFF
--- a/src/config/language.c
+++ b/src/config/language.c
@@ -125,33 +125,32 @@ void CONFIG_ReadLang(u8 idx)
         printf("Couldn't open language file: %s\n", file);
         return;
     }
-    while (fgets(line, sizeof(line), fh) != NULL) {
-        u16 hash;
-        if(line[0] == ':') {
-            CLOCK_ResetWatchdog();
-            fix_crlf(line+1);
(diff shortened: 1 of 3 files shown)


howard0su:
Need someone testing on devo8. So far devo7e and devof12e doesn't support localization, which are the only TXs I have.

This change is another step towards full localization support for f12e.

howard0su:
Add back the support for old files. As we are sorting the strings in new format by hash. I assume the first hash byte is not ':' in new format. So I can avoid to add version number here. I verified that all of current languages works fine in emu_devo8.

howard0su:
Can I get review on this change? Some additional work to create font file only contains the characters in localized string file depends on this.

howard0su:
Need one more approve.

PR #512, opened 2018-12-24
Use ASCII only font for devo7e

howard0su:
No localization support for devo7e. So generate a ascii (0-255) only font to reduce the size of ROM and the time to flash a devo7e.


howard0su:
this may applies to at9 15normal font as well. I don't have a at9 and don't know if there is a plan to support location for at9.

howard0su:
Can I get review on this change? Some additional work to create font file only contains the characters in localized string file depends on this.

I am new to Perl. Please review Perl change including coding style and naming.

hexfet:
Shouldn't this apply to all transmitters with NO_LANGUAGE_SUPPORT defined in target_defs.h?  And that means it should not apply to devo7e-256.

I'm not a perl person either but those changes look okay to me.

howard0su:
@hexfet you are right. I will revert the change for devo7e-256.

I think you made a good point that we need support a better customization system. In today's implementation, we cannot do the different build tasks based on the different definition in target_defs.h. I will start an issue to capture the work.

howard0su:
this is not required for some Tx which is not including 12/15normal.fon anyway.

TheRealMoeder:
How much Rom is there actually to gain? I was always thinking that we could possibly provide localized 7e builds by preprocessing lang files in the build process

howard0su:
Please hold, I am working on supporting localization for devo7e.

TheRealMoeder:
@howard0su shall we close this?

PR #513, opened 2018-12-24
Change the default font size on AT9 to 15p

howard0su:
Based on the change of generate ascii font file, we can pack 15normal.fon into the 64KB SPI flash on AT9. So switch it to 15p as default which is much more readable.


howard0su:
No hardware to test. only verify on at9 emulator.

howard0su:
@hexfet if you like this change, I can cleanup this PR.

howard0su:
@hexfet can I get a review and testing on real Tx?

I cleanup this PR to remove the dependency on #512. Let's get this in if this is proper change. #512 may take a while to see if we can get localization support on limit ROM.

hexfet:
I don't have an at9 to test with.  It's probably best to go ahead with it and address any feedback if it comes.

PhracturedBlue:
I don't think there is any reason to wait for someone with an AT9 before making this change.  Go ahead and merge it.

PR #517, opened 2018-12-25
Fix issue with exiting USB mode in t8sg V2 transmitters. 

hexfet:
After powering up in USB mode, the next press of the power button would cause immediate reset on T8SG V2 transmitters.  The CheckPowerSwitch function assumed it was being called every 5ms from the event loop.  That's not true after powering up in USB mode.  Changed the function to be independent of call rate.


howard0su:
If the main loop is not every 5ms anymore, how about the handling of stick data? Should we stop the transmission when USB is ON?

hexfet:
There's no main loop running in this state.  The code is in an infinite loop in USB_Connect from line 55 of main.c.  Only way to exit this state is through the power switch.  Only the T8SG V2 transmitters have the momentary switch that needs the "debounce" timing.

Issue #516, opened 2018-12-25
need rerun make language


howard0su:
we don't have a way to enforce the code change which impact localization need run 'make language'. current master has many loc string changes already.

PhracturedBlue:
Fixed by #629  which will automatically upload string changes to transifex and download and build them in the nightlies

PR #520, opened 2018-12-27
regenerate language file for loc

howard0su:
We should not commit this as it is. Several new reference strings are too similar with the existing ones. we should fix it. Work for #516.


howard0su:
close this as @PhracturedBlue will handle this.

Issue #484, opened 2018-12-14
Some protocols doesn't work in devo7e (modular build)


howard0su:
So far I notice:
1. CRSF
2. SBUS
3. WFLY

TheRealMoeder:
See PR #487

howard0su:
@goebish did u verify wily works in modular build?

goebish:
I don't see why it wouldn't work in modular build only, it does nothing particular.

howard0su:
One guess is that wfly protocol did not pick the channels based on rssi. On my devo7e the tax power is only 10mw and my WiFi is much stronger. It cause receiver cannot gets the signal. Possible?

goebish:
None of the protocols pick up the frequency based on RSSI, I don't think that's the issue.
I don't have a WFLY receiver nor a 7e, just for testing, can you try to set MAX_WAIT_WRITE to 100 here ?
https://github.com/DeviationTX/deviation/blob/master/src/protocol/wfly_cyrf6936.c#L198

howard0su:
I tested to set it to 100. Doesn't work still. and I found more serious problem with WFLY (or something else).

Here repro steps:
1. flash zip package to devo7e.
2. configure the model1 to use WFLY as protocol.
3. run 'Bind' twice. (no receiver so it will never success). First bug: the wait time is so big. "Press ENT to stop 1492 seconds left".
4. Exit to main menu and select channel monitor. you will find only ch1 has the input when you move two sticks. I found it is caused model1.ini is somehow corrupted.

The model1.ini for your reference:
[https://pastebin.com/raw/YFyjcjAz](url)

howard0su:
More information from error.txt. Any idea?
[Soft fault]
devo7e-v5.0.0-ebc99d2
R0 = 40003800
R1 = 00000083
R2 = 000005dc
R3 = 00000002
R12 = 0000ffff
LR [R14] (subroutine call return address) = 200045fd
PC [R15] (program counter) = 0801bfdc
PSR = 61000000
BFAR = e000ed38
CFSR = 00000000
HFSR = 00000000
DFSR = 00000000
AFSR = 00000000
SCB_SHCSR = 00000000
Top of Stack:20004000
Stack Detect:20003f20
Backtrace:
20003f7c : 0800d8c7
20003f84 : 0800e2d5
20003f94 : 0800e36f
20003f98 : 0800e34d
20003fa4 : 0800929b
20003fbc : 08009375
20003fdc : 08009d93
[...]

PR #521, opened 2018-12-28
Fix regression on channel mapping

howard0su:
Channel mapping doesn't work expected in some protocol modules due to the variables are not initialized proper in the modular. Use a number instead of pointer to return from modular and convert it to the real pointer in main firmware.


howard0su:
Fix issue #484 thank @goebish to point out the problem.

PR #522, opened 2018-12-28
Generate a mini fon file

howard0su:
A mini font file only contains the font data for the characters which exist in localization files. Also support the filter out all ascii characters (<255), which F12E supports via its Font ROM.

The change is not ready to check-in. it depends on #511 #512.


howard0su:
The current format of font file doesn't work very well for mini font file. It can have too many ranges (285 ranges as it is today.) I am thinking to introduce a different format to work around this problems.

1. Just use a .H to include font into the main firmware. this is a good option for f12e as it only has 64KB SPI flash. And SPI flash needs store localization files.
2. A font file which list all the characters in the first level with the unicode and offset.

thoughts?

PR #523, opened 2018-12-29
two buffer is not needed be static to save 16 bytes RAM

howard0su:


src/extended_audio.c
```diff
@@ -132,9 +134,8 @@ int AUDIO_Play(u16 music) {
       break;
     }
     case AUDIO_DF_PLAYER: {
```
hexfet: Brace block no longer needed here.

src/extended_audio.c
```diff
@@ -165,9 +166,8 @@ void AUDIO_SetVolume() {
       case AUDIO_AUDIOFX: // AUDIOFX only allows up down selection of volume...not implemented
       case AUDIO_NONE: break;	// Play beeps...
       case AUDIO_DF_PLAYER: {
```
hexfet: Brace block no longer needed.

hexfet:
Now the ROM and RAM usage is the same as master. No reason for the change.

howard0su:
Interesting, ROM size increased a bit more. use devo8 as example: RAM do decrease 16bytes, but ROM increased a bit more.
Master:
ROM: 0x08004000 - 0x0803dca8 = 231.16kB
RAM: 0x20000000 - 0x20009010 =  36.02kB

After:
ROM: 0x08004000 - 0x0803dcd4 = 231.21kB
RAM: 0x20000000 - 0x20009000 =  36.00kB

hexfet:
Interesting.  I had compared the 7e and they are the same.  Not sure whether RAM or ROM is more precious in the non-modular builds.  Approving because I like this style better.

devo7e:
master:
ROM: 0x08003000 - 0x0801fdb0 = 115.42kB
RAM: 0x20000000 - 0x200028a0 =  10.16kB
ram_save:
ROM: 0x08003000 - 0x0801fdb0 = 115.42kB
RAM: 0x20000000 - 0x200028a0 =  10.16kB

howard0su:
because the changed code only applies to extended audio. Devo7e does not support it.

howard0su:
okay, go back with one buffer approach. use 6 bytes RAM to save 36 bytes ROM.

PR #529, opened 2019-01-02
Reduce ROM usage of music module

howard0su:
1. Parse the name of tone instead of a table.
2. Only store one scale of freq and caculate others.

### DIFF
--- a/src/music.c
+++ b/src/music.c
@@ -24,28 +24,6 @@ static struct {u8 note; u8 duration;} Notes[100];
 static u8 Volume;
 static u8 next_note;
 static u8 num_notes;
-struct NoteMap {
-    const char str[4];
-    u16 note;
-};
-static const struct NoteMap note_map[] = {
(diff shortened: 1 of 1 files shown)


howard0su:
Don't know how to test this. The ROM size saving is noticable:

Before:
ROM: 0x08003000 - 0x0801fdb0 = 115.42kB
RAM: 0x20000000 - 0x200028a0 =  10.16kB

After:
ROM: 0x08003000 - 0x0801fc78 = 115.12kB
RAM: 0x20000000 - 0x200028a0 =  10.16kB

howard0su:
verified on 7e, at least startup music works fine.

howard0su:
maybe we should have a longer music for startup. it is fine to compose music like that.

TheRealMoeder:
> maybe we should have a longer music for startup. it is fine to compose music like that.

Every user is open to do so. Feel free to compose a deviation theme if you wish ;-)

PR #531, opened 2019-01-02
Proper use HAS_TOUCH macro

howard0su:


src/std.h
```diff
@@ -68,7 +68,7 @@
     #if ! defined BUILDTYPE_DEV  && ! DEBUG_WINDOW_SIZE
         //Use this instead of printf(args...) because this will avoid
         //compile warnings
-        #define printf if(0) tfp_printf
+        #define printf while(0) tfp_printf
```
hexfet: What's the reason for this change?

hexfet:
What are your thoughts on the runtime usage of HAS_TOUCH in these files?
pages/320x240x16/main_page.c
pages/320x240x16/model_loadsave.c
pages/320x240x16/pages.c
pages/320x240x16/tx_configure.c
pages/common/_main_config.c
pages/common/_main_page.c
target/common/devo/spi_touch.c
timer.c

howard0su:
you made right point. but those places the compiler already optmized out the dead code. I propose this is because SPITouch_Init somehow shows up in .map file as a function.

PhracturedBlue:
I don't really love this.  I specifically went through the code at one point and replaced #ifdefs with if()  wherever possible (I think this had to do with having the compiler do syntax checking, but it was a long time ago).  The compiler should be optimizing out this code already as it does elsewhere.  It would be worth understanding why this case is special.

howard0su:
okay. make sense.

PR #533, opened 2019-01-03
Add Scanner support to 128x64x1 screen

howard0su:
The freq scanner is ported over to 128x64 screen Tx including devo10/devo12e/t8sg/devo7e-256.


howard0su:
I don't have real hardware to test. please help.

PR #534, opened 2019-01-04
Fix perl build on Windows

howard0su:


howard0su:
@goebish reported that perl on windows doesn't support "S<".

PhracturedBlue:
You probably know better than I do.  Is it worth using 2 bytes and manually specifying the order?

howard0su:
Yes, have to. otherwise lang.* cannot fit into 64kb flash with other media.

hexfet:
The byte order modifier works on my Windows perl 5.1.2.  Doesn't make sense a feature like that would be OS dependent.  I don't build deviation on windows though.  What was the failure?

Wouldn't this change mean the file would be different if build on a big-endian machine?  Which may never happen - don't have one myself to test on.

goebish:
It doesn't work on my side, Windows, perl 5.8.8 msys-int64
https://github.com/DeviationTX/deviation/commit/5b3d4214f7cf65d71be3a9ef0e8cd10cabc9ee2e#commitcomment-31830468

PhracturedBlue:
but it should probably be 'CC' (with appropriate casting) instead of 'S' to ensure the endianness?

PR #539, opened 2019-01-04
optimize module load code

howard0su:
1. Simplify the load logic to read .mod file
2. Fix error handling
3. Make sure we switch back font before we disable dialog.
4. Save about 50 bytes ROM.

### DIFF
--- a/src/protocol/protocol.c
+++ b/src/protocol/protocol.c
@@ -129,9 +129,10 @@ void PROTOCOL_Load(int no_dlg)
 {
     (void)no_dlg;
 #ifdef ENABLE_MODULAR
-    *loaded_protocol = 0;
-    if(! PROTOCOL_HasModule(Model.protocol))
+    if(! PROTOCOL_HasModule(Model.protocol)) {
+        *loaded_protocol = 0;
         return;
+    }
     if(*loaded_protocol == Model.protocol)
         return;
     char file[25];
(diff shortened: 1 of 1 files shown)


src/protocol/protocol.c
```diff
@@ -129,18 +129,17 @@ void PROTOCOL_Load(int no_dlg)
 {
     (void)no_dlg;
 #ifdef ENABLE_MODULAR
-    if(! PROTOCOL_HasModule(Model.protocol)) {
-        *loaded_protocol = 0;
+    *loaded_protocol = 0;
+    if(! PROTOCOL_HasModule(Model.protocol))
         return;
```
PhracturedBlue: The reason the code didn't do this originally is that we don't want to invalidate the protocol unless we are actually changing it.

src/protocol/protocol.c
```diff
@@ -159,36 +158,23 @@ void PROTOCOL_Load(int no_dlg)
...
-        if(len) {
-            memcpy(ptr, buf, len);
-            ptr += len;
-        }
-        size += len;
-        if (len != 256)
-            break;
-    }
+    fread(loaded_protocol, 1, 4 * 1024, fh);
     fclose(fh);
```
PhracturedBlue: I think this is ok, but need to double check that all implementations of fread() can handle large reads.  there may be some FS issue here.  Probably not though.

howard0su: I confirm both pf_read and df_read support read more than more the sector size.

PR #546, opened 2019-01-06
Fix trim calculation

PhracturedBlue:
The trim code is basically broken for many cases.
It is supposed to work as follows:
```
trim step moves in 0.1% increments from 0% to 10%
trim step moves in 1% increments from 10% to 100%
trim value has a maximum range of -100% to 100% and is independent of the step.  It should move 1% for a short-press of the trim button and 10% for a long-press of the trim button
the channel value should be adjusted based on trim-step * trim-value
```
The current code does not calculate the channel-adjustment properly with trim-step > 1% and does not consistently update the step-size with trim-step > 1%

This patch affects how trims are calculated and so could theoretically break models, but the trims were basically unusable in the range that the patch fixes anyway.

### DIFF
--- a/src/mixer.c
+++ b/src/mixer.c
@@ -448,6 +448,7 @@ s32 MIXER_ApplyLimits(unsigned channel, struct Limit *limit, volatile s32 *_raw,
     return value;
 }
 
+//Returns the raw trim value (between -100 and 100)
 s8 *MIXER_GetTrim(unsigned i)
 {
     if (Model.trims[i].sw) {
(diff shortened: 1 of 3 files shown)


src/mixer.c
```diff
@@ -448,6 +448,7 @@ s32 MIXER_ApplyLimits(unsigned channel, struct Limit *limit, volatile s32 *_raw,
     return value;
 }
 
+//Returns the raw trim value (between -100 and 100)
```
howard0su: why not add a test case to cover this?

PhracturedBlue:
I am in the process of writing full test coverage for the mixer.  That is what led me to find that the trim was broken in the 1st place.  So yes, I will be adding test coverage to ensure that the trims work properly, just not finished yet

PR #547, opened 2019-01-06
Save 128 bytes for 7e in mixer.c

TheRealMoeder:


src/mixer.c
```diff
@@ -312,38 +312,39 @@ void MIXER_ApplyMixer(struct Mixer *mixer, volatile s32 *raw, s32 *orig_value)
...
             //value initially represents 20ths of seconds to cover 60-degrees
             //convert value to represent #msecs to cover 60-degrees (zero->full)
             if (value == 0 || orig_value == NULL) {
-                value = raw[mixer->dest + NUM_INPUTS + 1];
+                value = scaled_value;
                 break;
             }
-            value = abs(RANGE_TO_PCT(value)) * 50; //'50' represents 1/20 of a second (1000msec/20 = 50)
+            value = abs(RANGE_TO_PCT(value)) * 50;
             //rate represents the maximum travel per iteration (once per MEDIUM_PRIORITY_MSEC)
```
PhracturedBlue: Please restore the comment.  I put it there because I couldn't recall where the 50 came from in the 1st place

src/mixer.c
```diff
@@ -353,36 +354,34 @@ void MIXER_ApplyMixer(struct Mixer *mixer, volatile s32 *raw, s32 *orig_value)
...
-                if ((value > *orig_value && value < new_value) ||
-                    (value < *orig_value && value > new_value) ||
-                    (abs(value - new_value) <= 10)) {
+                if ((value > *orig_value && value < scaled_value) ||
+                    (value < *orig_value && value > scaled_value) ||
+                    (abs(value - scaled_value) <= 10)) {
                     MIXER_SET_BEEP_LOCK(mixer, 1);
-                    MUSIC_Play(MUSIC_MAXLEN);
+                    MUSIC_Play(MUSIC_SAVING);
                 }
```
PhracturedBlue: This change of song doesn't seem related to the rest of the patch...why is it here?

PhracturedBlue:
I was ok with the 1st iteration (50 bytes) since it modified code that I had already written tests for.  The new version modifies the cyclic which I haven't written tests for.  I'm sure it is fine, but would prefer to defer cleanups of the mixer on code that does not yet have functional tests

PhracturedBlue:
also, can you rebase to the latest code.  We should now actually see how many bytes are saved in the travis builds (I hope)

TheRealMoeder:
> I was ok with the 1st iteration (50 bytes) since it modified code that I had already written tests for. The new version modifies the cyclic which I haven't written tests for. I'm sure it is fine, but would prefer to defer cleanups of the mixer on code that does not yet have functional tests

Then let's just wait with the merge. We're not in the need for memory right now anyways, as we already have 1.5 kB spared for which we will have to decide what to do with it ;) I just fun to do look for opportunities 😸

TheRealMoeder:
> also, can you rebase to the latest code. We should now actually see how many bytes are saved in the travis builds (I hope)

It works, but delta for RAM is wrong, it calculates as rOm_old - rAm_new currently

PhracturedBlue:
Looks like the delta calculation is good for the ROM but not for the RAM...thought I fixed that already. Will fix. Still it is a pretty cool metric to see how a patch impacts all targets.
Given this new data, I could trigger a failure if the memory increased a certain percentage (it would really just be a warning that the patch has significant impact).  Thoughts?

TheRealMoeder:
> Looks like the delta calculation is good for the ROM but not for the RAM...thought I fixed that already. Will fix. Still it is a pretty cool metric to see how a patch impacts all targets.
> Given this new data, I could trigger a failure if the memory increased a certain percentage (it would really just be a warning that the patch has significant impact). Thoughts?

line 212 is wrong

TheRealMoeder:
I will update this when PR #555 is merged.

TheRealMoeder:
@PhracturedBlue: after rebase and resolving conflicts mixer tests passed. Merge?

PR #548, opened 2019-01-06
Store build image size (ram/rom) in GitHub status

PhracturedBlue:
This will preserve image size for each build and annotate it (along with deltas) onto pull requests

### DIFF
--- a/utils/archive_build_size.py
+++ b/utils/archive_build_size.py
@@ -1,0 +1,216 @@
+#!/usr/bin/env python
+"""Track changes in the size of an image using Travis-CI and GitHub statuses.
+
+This code is adapted from https://github.com/danvk/travis-weigh-in
+
+Usage:
+    archive_build_size.py <MAKETARGET>
+
+This should be run inside of a Travis-CI worker.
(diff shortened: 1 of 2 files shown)


utils/archive_build_size.py
```diff
@@ -0,0 +1,216 @@
...
+    headers = {'Authorization': 'token ' + GITHUB_TOKEN}
+    request = urllib2.Request(url, None, headers)
+    res = urllib2.urlopen(request)
+    raise_for_status(url, res)
+    return json.loads(res.read())
+
+
+def get_context(filename):
+    """Define status context"""
+    return "image_size/" + filename
```
TheRealMoeder: Can we shorten image_size so the status does not get truncated in pull requests overview?

howard0su: Agree. We can display delta first then the size. So we can at least see the diff without mouse hover. We can remove unit b as well.

PhracturedBlue: I removed the 'image_size/' part and the 'b'.  That seems to provide enough space

PR #551, opened 2019-01-06
Shrink voice ROM usage

TheRealMoeder:
To make it fit devo7e if deemed useful


TheRealMoeder:
That was some really nasty code I wrote back then....

PR #554, opened 2019-01-08
Strip RAM needs for voice

TheRealMoeder:
Only read labels when inside voiceconfig menu by rerunning CONFIG_VoiceParse

### DIFF
--- a/src/config/voice.c
+++ b/src/config/voice.c
@@ -32,6 +32,7 @@ const char SECTION_VOICE_CUSTOM[] = "custom";
 
 static int ini_handler(void* user, const char* section, const char* name, const char* value)
 {
+    // ini_handler returns a requested mp3 label passed at *user to tempstring
     u16 req_id = *((long*)user);
     u16 id = atoi(name);
     const char* ptr = value;
@@ -49,7 +50,7 @@ static int ini_handler(void* user, const char* section, const char* name, const 
 #if HAS_MUSIC_CONFIG
     if ( k && (req_id != MAX_VOICEMAP_ENTRIES) && (req_id == id) ) {
         strlcpy(tempstring, value, k+1);
-        tempstring[MAX_VOICE_LABEL] = '\0';
+        tempstring[MAX_VOICE_LABEL] = '\0'; // limit label length
         return 1;
     }
 #endif
@@ -103,7 +104,7 @@ const char* CONFIG_VoiceParse(unsigned id)
     }
     if (id >= CUSTOM_ALARM_ID) {
         if (CONFIG_IniParse(filename, ini_handler, &id)) {
-            //ini handler will set tempstring to label of mp3id
+            //ini handler will return tempstring with label of id
         }
     }
     return tempstring;
--- a/src/config/voice.c
+++ b/src/config/voice.c
@@ -102,7 +102,7 @@ const char* CONFIG_VoiceParse(unsigned id)
             strcpy(tempstring, "");
         }
     }
-    if (id >= CUSTOM_ALARM_ID) {
+    if ( (id < MAX_VOICEMAP_ENTRIES) && (id >= CUSTOM_ALARM_ID) ) {
         if (CONFIG_IniParse(filename, ini_handler, &id)) {
             //ini handler will return tempstring with label of id
         }
--- a/src/config/voice.c
+++ b/src/config/voice.c
@@ -52,33 +52,30 @@ static int ini_handler(void* user, const char* section, const char* name, const 
         return 1;
     }
 #endif
-
-    if ( (*req_id == MAX_VOICEMAP_ENTRIES) && MATCH_SECTION(SECTION_VOICE_GLOBAL) ) {
-        for (int i = 0; i < CUSTOM_ALARM_ID; i++) {
-            snprintf(tempstring, 4, "%d", i);
-            if (MATCH_KEY(tempstring)) {
-                voice_map[i].duration = duration;
-                voice_map[i].id = i;
-                printf("id %d, duration %d\n", id, duration);
-                return 1;
+    if ( *req_id == MAX_VOICEMAP_ENTRIES ) {
+        if (MATCH_SECTION(SECTION_VOICE_GLOBAL)) {
+            for (int i = 0; i < CUSTOM_ALARM_ID; i++) {
+                snprintf(tempstring, 4, "%d", i);
+                if (MATCH_KEY(tempstring)) {
+                    voice_map[i].duration = duration;
+                    voice_map[i].id = i;
+                    return 1;
+                }
             }
         }
-    }
-    if ( (*req_id == MAX_VOICEMAP_ENTRIES) && MATCH_SECTION(SECTION_VOICE_CUSTOM) ) {
-        voice_map[voice_map_entries].duration = duration;
-        voice_map[voice_map_entries].id = atoi(name);
-        voice_map_entries++;
-        return 1;
-    }
-    if ( *req_id == MAX_VOICEMAP_ENTRIES ) {
+        if (MATCH_SECTION(SECTION_VOICE_CUSTOM)) {
+            voice_map[voice_map_entries].duration = duration;
+            voice_map[voice_map_entries].id = atoi(name);
+            voice_map_entries++;
+            return 1;
+        }
         printf("Unknown entry in voice.ini: %s\n", value);
         return 0;
     }
-
     return 1; // voice label ignored
 }
 
-void CONFIG_VoiceParse(u16 music)
+void CONFIG_VoiceParse(u16 id)
 {
     #ifdef _DEVO12_TARGET_H_
     static char filename[] = "media/voice.ini\0\0\0"; // placeholder for longer folder name
@@ -95,16 +92,15 @@ void CONFIG_VoiceParse(u16 music)
     #else
     char filename[] = "media/voice.ini";
     #endif
-    if (music == MAX_VOICEMAP_ENTRIES) { // initial parse of voice.ini
+    if (id == MAX_VOICEMAP_ENTRIES) { // initial parse of voice.ini
         voice_map_entries = CUSTOM_ALARM_ID; // Reserve space in map for global alerts
-        if (CONFIG_IniParse(filename, ini_handler, &music)) {
+        if (CONFIG_IniParse(filename, ini_handler, &id)) {
             printf("Failed to parse voice.ini\n");
             Transmitter.audio_player = AUDIO_NONE; // disable external voice output
         }
     }
-    if (music >= CUSTOM_ALARM_ID) {
-        if (CONFIG_IniParse(filename, ini_handler, &music)) {
-            printf("Failed to find label in voice.ini\n");
+    if (id >= CUSTOM_ALARM_ID) {
+        if (CONFIG_IniParse(filename, ini_handler, &id)) {
         }
     }
 }
--- a/src/config/voice.c
+++ b/src/config/voice.c
@@ -48,7 +48,7 @@ static int ini_handler(void* user, const char* section, const char* name, const 
 
 #if HAS_MUSIC_CONFIG
     if ( (*req_id != MAX_VOICEMAP_ENTRIES) && (*req_id == id) ) {
-        strlcpy(voicelabel,getfield(tmp, 1), MAX_VOICE_LABEL);
+        strlcpy(tempstring,getfield(tmp, 1), MAX_VOICE_LABEL);
         return 1;
     }
 #endif
--- a/src/config/voice.c
+++ b/src/config/voice.c
@@ -18,6 +18,7 @@
 #include "common.h"
 #include "model.h"
 #include "../music.h"
+#include "extended_audio.h"
 #include "tx.h"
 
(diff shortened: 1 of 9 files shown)


src/config/voice.c
```diff
@@ -39,44 +40,43 @@ static const char* getfield(char* value, int num) {
 
 static int ini_handler(void* user, const char* section, const char* name, const char* value)
 {
-    (void) user;
+    u16 *req_id = user;
```
howard0su: Why not fetch the value here?

src/config/voice.c
```diff
@@ -39,44 +40,43 @@ static const char* getfield(char* value, int num) {
...
     char tmp[100];
```
howard0su: Can we use tempstring？

src/music.h
```diff
@@ -112,7 +112,7 @@ struct VoiceMap {
     u16 id;
     u16 duration;
 #if HAS_MUSIC_CONFIG
-    char label[MAX_VOICE_LABEL];
+//    char label[MAX_VOICE_LABEL];
 #endif
```
howard0su: Remove this instead of comment out

src/config/voice.c
```diff
@@ -39,44 +40,43 @@ static const char* getfield(char* value, int num) {
...
     strlcpy(tmp, value, 100);
     int duration = atoi(getfield(tmp,2));
-#if HAS_MUSIC_CONFIG
-    char label[MAX_VOICE_LABEL];
-    strlcpy(label,getfield(tmp, 1), MAX_VOICE_LABEL);
-#endif
+    u16 id = atoi(name);
```
PhracturedBlue: this is the only place in all of deviation that strtok() is called.  Is there an alternative that wouldn't require including it (or the strcpy that goes with modifying the string)?

src/config/voice.c
```diff
@@ -39,44 +40,43 @@ static const char* getfield(char* value, int num) {
...
-        return 1;
-    }
-    printf("Unknown entry in voice.ini: %s\n", value);
-    return 0;
+    return 1; // voice label ignored
 }
 
-void CONFIG_VoiceParse()
+void CONFIG_VoiceParse(u16 id)
 {
```
PhracturedBlue: why use a u16 here?  It is almost always beter to use 'unsigned' for passed variables (and only use the reduced size variables for global/static

src/pages/common/_voiceconfig_page.c
```diff
@@ -122,26 +122,36 @@ static const char *voicelbl_cb(guiObject_t *obj, const void *data)
...
+            }
             break;
 #if NUM_AUX_KNOBS
         case VOICE_SRC_AUX:
-            if(Model.voice.aux[idx - VOICE_SRC_AUX].music)
-                return voice_map[Model.voice.aux[idx - VOICE_SRC_AUX].music].label;
+            if(Model.voice.aux[idx - VOICE_SRC_AUX].music) {
+                CONFIG_VoiceParse(voice_map[Model.voice.aux[idx - VOICE_SRC_AUX].music].id);
+                return tempstring;
+            }
```
PhracturedBlue: I really don't like this.  A function should not use tempstring as an implicit argument/return passing  If you need to pass it, have CONFIG_VoiceParse return tempstring, and here just run retun CONFIG_VoiceParse(...)

src/config/voice.c
TheRealMoeder: removed stroke usage and used temp string

src/pages/common/_voiceconfig_page.c
TheRealMoeder: I hope the solution is better now. It is definitely smaller ;)

TheRealMoeder:
There is still a little mess concerning the naming of this feature, as it was moved from extended audio to voice in the process of development after a discussion on the forum. I will address this issue in a future pull request, which will also enhance the menu features.

TheRealMoeder:
One thing I noticed is, that the deltas are wrong if the pr branch isn't rebased on the current master. I'm not sure there really is a solution to this, but it is a little misleading

src/config/voice.c
```diff
@@ -28,55 +30,54 @@
...
-    if (MATCH_SECTION(SECTION_VOICE_GLOBAL)) {
-        for (int i = 0; i < CUSTOM_ALARM_ID; i++) {
-            snprintf(tempstring, 4, "%d", i);
-            if (MATCH_KEY(tempstring)) {
-                voice_map[i].duration = duration;
-                voice_map[i].id = i;
 #if HAS_MUSIC_CONFIG
-                strcpy(voice_map[i].label, label);
+    if ( k && (req_id != MAX_VOICEMAP_ENTRIES) && (req_id == id) ) {
+        strlcpy(tempstring, value, k+1);
```
howard0su: still implict return tempstring?

TheRealMoeder: I know, but ini handler can only return into. How else to solve this problem? If we don't rely on ini.c code, rom size will likely grow quite a bit...

howard0su: Can u add some comment to make the logic explicit? I am fine with that.

TheRealMoeder:
Please squash when merging

howard0su:
is this change ready? I may touch the same file in another change. it will be better that you merge this up front.

TheRealMoeder:
I actually was thinking of removing voice map and parsing voice.ini on demand just like we do for sound.ini...haven't gotten to it yet though

TheRealMoeder:
But we might as well merge this (it works fine for me on t8sg) and I'll work on the rest when you have your model.c changes ready

PhracturedBlue:
some of these lint errors are interesting.
Do not fix:
```Use int16/int64/etc, rather than the C type long```
I don't think that is portable with pointers
```Almost always, snprintf is better than strcpy```
Certainly not true for us.  strlcpy is preferred to strcpy, but in this trivial case, maybe `tempstring[0] = 0`
```If you can, use sizeof(tempstring) instead of 4 as the 2nd arg to snprintf.```
I am not sure if this is a good rule in general.  It isn't a good rule for his case though

File: src/pages/320x240x16/voiceconfig_page.c [hunk 8]
```diff
@@ -32,12 +32,10 @@ enum {
...
     (void)data;
-        int row = y;
-        int voice_num = absrow;
         //Row 1
-        GUI_CreateLabelBox(&gui->name[relrow], COL1, row, COL2-COL1, 18, &LABEL_FONT,voiceconfig_str_cb, NULL, (void *)(long)voice_num);
-        GUI_CreateTextSelect(&gui->voiceidx[relrow],  COL2, row, TEXTSELECT_64, voice_test_cb, voiceid_cb, (void *)(long)voice_num);
-        GUI_CreateLabelBox(&gui->voicelbl[relrow], COL3, row, LCD_WIDTH - COL3 - ARROW_WIDTH, 18, &TINY_FONT, voicelbl_cb, NULL, (void *)(long)voice_num);
+        GUI_CreateLabelBox(&gui->name[relrow], COL1, y, COL2-COL1, 18, &LABEL_FONT, voiceconfig_str_cb, NULL, (void *)(long)absrow);
+        GUI_CreateTextSelect(&gui->voiceidx[relrow],  COL2, y, TEXTSELECT_64, voice_test_cb, voiceid_cb, (void *)(long)absrow);
+        GUI_CreateLabelBox(&gui->voicelbl[relrow], COL3, y, LCD_WIDTH - COL3 - ARROW_WIDTH, 18, &TINY_FONT, voicelbl_cb, NULL, (void *)(long)absrow);
```
PhracturedBlue: Lint Error:
```
36:  Use int16/int64/etc, rather than the C type long  
37:  Use int16/int64/etc, rather than the C type long  
38:  Use int16/int64/etc, rather than the C type long  
```

src/config/voice.c
```diff
@@ -28,55 +30,55 @@
...
-    (void) user;
-    char tmp[100];
-    strlcpy(tmp, value, 100);
-    int duration = atoi(getfield(tmp,2));
-#if HAS_MUSIC_CONFIG
-    char label[MAX_VOICE_LABEL];
-    strlcpy(label,getfield(tmp, 1), MAX_VOICE_LABEL);
-#endif
+    // ini_handler returns a requested mp3 label passed at *user to tempstring
+    u16 req_id = *((long*)user);
```
PhracturedBlue: Lint Error:
```
36:  Use int16/int64/etc, rather than the C type long  
```

src/config/voice.c
```diff
@@ -28,55 +30,55 @@
...
-        for (int i = 0; i < CUSTOM_ALARM_ID; i++) {
-            snprintf(tempstring, 4, "%d", i);
-            if (MATCH_KEY(tempstring)) {
-                voice_map[i].duration = duration;
-                voice_map[i].id = i;
 #if HAS_MUSIC_CONFIG
-                strcpy(voice_map[i].label, label);
+    if ( k && (req_id != MAX_VOICEMAP_ENTRIES) && (req_id == id) ) {
+        strlcpy(tempstring, value, k+1);
+        tempstring[MAX_VOICE_LABEL] = '\0'; // limit label length
```
PhracturedBlue: Lint Error:
```
53:  At least two spaces is best between code and comments  
```

src/config/voice.c
```diff
@@ -28,55 +30,55 @@
...
+        return 1;
+    }
 #endif
-                return 1;
+    if ( req_id == MAX_VOICEMAP_ENTRIES ) {
+        if (MATCH_SECTION(SECTION_VOICE_GLOBAL)) {
+            for (int i = 0; i < CUSTOM_ALARM_ID; i++) {
+                snprintf(tempstring, 4, "%d", i);
```
PhracturedBlue: Lint Error:
```
60:  If you can, use sizeof(tempstring) instead of 4 as the 2nd arg to snprintf.  
```

src/config/voice.c
```diff
@@ -28,55 +30,55 @@
...
-        voice_map[voice_map_entries].id = atoi(name);
-#if HAS_MUSIC_CONFIG
-        strcpy(voice_map[voice_map_entries].label, label);
-#endif
-        voice_map_entries++;
-        return 1;
-    }
-    printf("Unknown entry in voice.ini: %s\n", value);
-    return 0;
+    return 1; // voice label ignored
```
PhracturedBlue: Lint Error:
```
77:  At least two spaces is best between code and comments  
```

src/config/voice.c
```diff
@@ -92,9 +94,19 @@ void CONFIG_VoiceParse()
...
     #endif
-    if (CONFIG_IniParse(filename, ini_handler, NULL)) {
-        printf("Failed to parse voice.ini\n");
-        Transmitter.audio_player = AUDIO_NONE; // disable external voice output
+    if (id == MAX_VOICEMAP_ENTRIES) { // initial parse of voice.ini
+        voice_map_entries = CUSTOM_ALARM_ID; // Reserve space in map for global alerts
+        if (CONFIG_IniParse(filename, ini_handler, &id)) {
+            printf("Failed to parse voice.ini\n");
+            Transmitter.audio_player = AUDIO_NONE; // disable external voice output
+            strcpy(tempstring, "");
```
PhracturedBlue: Lint Error:
```
97:  At least two spaces is best between code and comments  
98:  At least two spaces is best between code and comments  
101:  At least two spaces is best between code and comments  
102:  Almost always, snprintf is better than strcpy  
```

src/config/voice.c
```diff
@@ -92,9 +94,19 @@ void CONFIG_VoiceParse()
...
+        }
+    }
+    if ( (id < MAX_VOICEMAP_ENTRIES) && (id >= CUSTOM_ALARM_ID) ) {
+        if (CONFIG_IniParse(filename, ini_handler, &id)) {
+            //ini handler will return tempstring with label of id
```
PhracturedBlue: Lint Error:
```
107:  Should have a space between // and comment  
```

src/pages/320x240x16/voiceconfig_page.c
PhracturedBlue: Lint Error:
```
36:  Use int16/int64/etc, rather than the C type long  
37:  Use int16/int64/etc, rather than the C type long  
38:  Use int16/int64/etc, rather than the C type long  
```

src/config/voice.c
PhracturedBlue: Lint Error:
```
36:  Use int16/int64/etc, rather than the C type long  
```

src/config/voice.c
```diff
@@ -28,55 +30,55 @@
...
+        strlcpy(tempstring, value, k+1);
+        tempstring[MAX_VOICE_LABEL] = '\0';  // limit label length
+        return 1;
+    }
 #endif
-                return 1;
+    if ( req_id == MAX_VOICEMAP_ENTRIES ) {
+        if (MATCH_SECTION(SECTION_VOICE_GLOBAL)) {
+            for (int i = 0; i < CUSTOM_ALARM_ID; i++) {
+                snprintf(tempstring, 4, "%d", i);
```
PhracturedBlue: Lint Error:
```
60:  If you can, use sizeof(tempstring) instead of 4 as the 2nd arg to snprintf.  
```

TheRealMoeder:
I fixed by your suggestions, merge?

PR #555, opened 2019-01-08
Only run cyclic calculation once.  Add several mixer tests

PhracturedBlue:
For some reason we were calling MIXER_CreateCyclicOutput() for each cyclic channel even though it does the full caclulation each time it is called.


src/mixer.c
```diff
@@ -178,13 +178,14 @@ void MIXER_CalcChannels()
     MIXER_EvalMixers(raw);
 
     //4th step: apply auto-templates
+    s32 cyc[3];
```
howard0su: don't like this naming. can we just use cyclic?

PhracturedBlue:
FYI, please don't merge this yet.  I will merge it when I am ready

PR #558, opened 2019-01-08
Add language tests

howard0su:
1. Add test case for language testing
2. s/NO_LANGUAGE_SUPPORT/SUPPORT_MULTI_LANGUAGE/
3. Add target_all.h to give default values to knob
4. Avoid include target_defs.h directly, always include target.h
5. mark some functions as static.


src/config/language.c
```diff
@@ -150,53 +149,83 @@ static void CONFIG_ReadLangV1(FILE* fh)
...
 #endif
 
-static void CONFIG_ReadLangV2(FILE* fh)
+static void ReadLangV2(FILE* fh)
 {
     u16 hash;
     unsigned pos = 0;
-    char line[MAX_LINE];
+    char tempstring[MAX_LINE];
     struct str_map *lookup = lookupmap;
```
PhracturedBlue: why is this here?  either use global tempstring or define a different variable, but please don't define one named 'tempsting'

howard0su: mismerge. want to use global tempstring to reduce stack usage.

PR #560, opened 2019-01-08
Add Screenshot based tests

howard0su:
1. Add LCD String tests
2. Add GUI label tests

Based on change #559


howard0su:
AssertScreenShot()
If there is no reference file under tests/320x240x8/, this function will print out a message and generate the file as reference. If the reference file is there, this function will perform a pixel to pixel compare. If failed, the test image will generate under filesyste/test for developer to checkout.

src/screen/320x240x16/lcd_string.c
```diff
@@ -295,3 +295,6 @@ void LCD_GetStringDimensions(const u8 *str, u16 *width, u16 *height) {
 void LCD_SetFontColor(u16 color) {
     cur_str.color = color;
 }
+
+#define TESTNAME string
+#include "tests.h"
```
PhracturedBlue: we probably shouldn't call this test function 'string'.  either 'lcd_string' or 'screen_string' or whatever

howard0su: how about call it text?

PhracturedBlue: I'd prefer 'font' or 'drawtext'

Issue #563, opened 2019-01-09
Now I can't build master with local Windows environment, why?


vladislavy:
$ make TARGET=devo8
make: flock: Command not found
make: *** [libopencm3/lib/libopencm3_stm32f1.a] Error 127

PR #565, opened 2019-01-09
Don't try to flock on Windows

howard0su:
There is no flock on Windows. So don't do flock on Windows.


howard0su:
Fix #563 @vladislavy

PR #569, opened 2019-01-10
Fix strict-alias of type FAT

howard0su:
In new compiler, strict-alias is assumed to have better optimization. The current implementation to use an array to simulate the struct FATFS breaks this assumption and cause gcc 8.2 generate wrong code. Fix it via using the exact type.

### DIFF
--- a/src/protocol/protocol.c
+++ b/src/protocol/protocol.c
@@ -131,6 +131,8 @@ void PROTOCOL_Load(int no_dlg)
     (void)no_dlg;
 #ifdef ENABLE_MODULAR
     FATFS ModuleFAT;
+    FILE *fh;
+
     if(! PROTOCOL_HasModule(Model.protocol)) {
         *loaded_protocol = 0;
         return;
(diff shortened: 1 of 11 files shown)


howard0su:
need test the new change

howard0su:
tested on:
- [ ] devof12e
- [ ] devo7e
- [ ] devo8
- [ ] devo10
- [ ] devo12

src/protocol/protocol.c
```diff
@@ -131,6 +130,9 @@ void PROTOCOL_Load(int no_dlg)
 {
     (void)no_dlg;
 #ifdef ENABLE_MODULAR
+    FATFS ModuleFAT;
+    FILE *fh;
+
     if(! PROTOCOL_HasModule(Model.protocol)) {
```
PhracturedBlue: This does add to the stack usage, right (compared to reusing the font struct).  It is probably not relevant, but do you have an idea of how our worst-case stack usage looks?

howard0su: Understand you concern. FAT structure for devo7e is pretty small 0x28 bytes. Our major stack consumer is still in mixer code. Check the issue I reported before #479. Since we have any callback, the script I use cannot really sum up the stack usage. But the report still gives u a bit idea on how much stack we are using.

PhracturedBlue: Just checking.  I have no real concern about it

howard0su: please merge. then I can start gcc8.2 testing.

howard0su:
need another code review and approve,

PR #572, opened 2019-01-12
Fix crash in non-localized Tx

howard0su:
This is regression when converting to new MULTI_LANGUAGE knob.


howard0su:
please merge this. this is a new regression introduced by me.

Issue #442, opened 2018-09-26
F12E-XMS : Power off during flight and recv. video (use FPV)


unmac:
Irregular Crash occurs when F12E-XMS (5.0.0). using FPV 
- version : 5.0.0-21674bd
- Bind model : cheerson cx95s
- protocol : DSM2
- tx mod.
  4 in 1 txmodule
  4MB flash up
- symptom
  Irregular power off while fly and recv. Video.
  (Create errors.txt)

errors.txt 1
'
[Soft fault]
devof12e-XMS-v5.0.0-21674bd
R0 = 00000045
R1 = 2000bf37
R2 = c0000000
R3 = 00000003
R12 = 2000bf30
LR [R14] (subroutine call return address) = 08009b15
PC [R15] (program counter) = 08009916
PSR = 21000000
BFAR = e000ed38
CFSR = 00000000
HFSR = 00000000
DFSR = 00000000
AFSR = 0000000`

unmac:
Some source code checks.
I suspect that the crash is caused by a watchdog.
I did not confirm the details.

In the TW8816_GetVideoStandard () or TW8816_SetVideoStandard () function
Delay occurrence is presumed to be the cause of watchdog.
(I2C approach ??)

I do not know if this symptom is only my TX.

Do not call VIDEO_GetStandard / VIDEO_SetStandard periodically in the VIDEO_Update () function, 
(to be call only when the value of videosrc changes), the   symptom does not occur.

There are three things that I would like to check further.
  - Adjustment of I2C timeout (LONG_TIMEOUT 0x10000, FLAG_TIMEOUT 0x1000)
  - reconfirmation of connection between I2C and LCD
  - Identify the location where the delay occurs

howard0su:
the analyze make sense to me. I will take a look.

howard0su:
unfortunately, GetStandard will not immediately return the correct value. We need defer the get/set standard after couple of seconds. will change this.

howard0su:
Can u test the build? 
https://www.deviationtx.com/downloads-new/category/1522-csujun-test-video

I don't have seeing such problem in my f12e. Please report back the result.

unmac:
ok. i will test the firmware.
I will report the result later.

unmac:
After about 2.5 hours of testing,
The symptom did not occur. (firmware : devof12e-XMS-v5.0.0-b360129)

I will use the test firmware few days

howard0su:
[...]

PR #573, opened 2019-01-12
Don't run video_setstandard in every event loop

howard0su:
Set a flag to let the _getstandard runs after video is enabled in 3 seconds.


howard0su:
Try to fix PR #442

howard0su:
please hold to wait for the feedback from community.

howard0su:
Consider merge per the feedback from the user in #442

PR #576, opened 2019-01-12
Fix the build of devof4

howard0su:
The logic to auto detect CRCOFFSET cannot work with the include. So let the target set LINKFILE to the file it is using directly.


PhracturedBlue:
I have an alternate solution which I think is cleaner

PR #578, opened 2019-01-12
Speedup travis builds

TheRealMoeder:
Some small changes to travis.yml to speed up builds a little:

1. git clone only shallow
2. omit apt-get update
3. git needs no update through apt
4. move away from docker container to trusty VM, as docker containers are phased out anyways by travis
5. set minimal language, as we only need a very basic environment

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ addons:
 #      - gcc-arm-none-eabi
 
 #Do NOT set these.  They set 'CC' which breaks the build
-#language: c
+language: minimal
 #compiler: gcc
 
 git:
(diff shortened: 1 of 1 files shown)


TheRealMoeder:
I also experimented with moving to the newer xenial environment, but this no longer includes mingw32 but mingw64, so some deeper changes would be required, for which I can‘t really see and advantages right now.

TheRealMoeder:
Total build time is about 25 % less at the current workload compared to master. But actually the gain is less, as the VM boots slower and this is not counted to the build time, so we cannot really measure it.

PhracturedBlue:
regardless of anything else we do, I see no reason not to speed up the build this way

TheRealMoeder:
Sorry, last commit had wrong syntax and interesting enough happened to stop travis working without and error message. I already pushend a commit to fix this.

PR #582, opened 2019-01-13
Fix calibration (issue #574)

### LINKED ISSUE #574: Calibration is broken in master
after calibration, the stick value and swich value will not change anymore. @TheRealMoeder @PhracturedBlue

hexfet:
Run mixer calculations on timer after de-initializing protocol.  Ensures channel data is updated for functions like calibration.  Fixes #574 .

### DIFF
--- a/src/protocol/protocol.c
+++ b/src/protocol/protocol.c
@@ -121,6 +121,7 @@ void PROTOCOL_DeInit()
     CLOCK_StopTimer();
     if(Model.protocol != PROTOCOL_NONE && PROTOCOL_LOADED)
         PROTO_Cmds(PROTOCMD_DEINIT);
+    CLOCK_StartMixer(); // run mixer on timer so channels are updated for things like calibration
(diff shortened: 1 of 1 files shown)


src/protocol/protocol.c
```diff
@@ -121,6 +121,7 @@ void PROTOCOL_DeInit()
     CLOCK_StopTimer();
     if(Model.protocol != PROTOCOL_NONE && PROTOCOL_LOADED)
         PROTO_Cmds(PROTOCMD_DEINIT);
+    CLOCK_StartMixer(); // run mixer on timer so channels are updated for things like calibration
```
howard0su: I prefer putting the comments above the line, which is more common practice and also make the width of line small.

PhracturedBlue: I actually prefer this way because it makes it more likely the comment will be preserved if the line is moved.  But it isn't a strong rule

PR #583, opened 2019-01-14
Reduce stack usage of MIXER_SetMixers by ~2200 bytes

PhracturedBlue:
Reimplement mixer-dependency code to not store a complete copy of Model.mixers in stack.
Reduces stack usage of this function from ~2588bytes to ~380 bytes.  Code cost is ~68 bytes

### DIFF
--- a/src/mixer.c
+++ b/src/mixer.c
@@ -579,77 +579,80 @@ int compact_mixers() {
     return i;
 }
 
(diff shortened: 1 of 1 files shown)


howard0su:
worth for that. Maybe devo7e has enough RAM now for localization consider stack saving and voice RAM saving.

TheRealMoeder:
Voice is not available in standard 7e builds, so no ram to gain there

PR #591, opened 2019-01-15
Hide permanent timer under knob

howard0su:


PhracturedBlue:
What is the point of this?  It doesn't save much in ram or rom

howard0su:
4bytes saving for devo7e purposely. :)

PR #592, opened 2019-01-15
move static variables into struct pagemem

howard0su:


howard0su:
i need figure out how to test permanent timer. I never use this feature before.

PR #595, opened 2019-01-15
Rearrange struct Model to save a little of ROM and RAM

howard0su:


howard0su:
seems not consistent in all Tx. some increases the others decrease.

PhracturedBlue:
In theory you can use pahole to see what is going on, but I've never bothered to recompile it to read arm binaries.  I generally run  it on the emu builds, and find it close enough as they are also 32bit w/ 4 byte alignment.

PR #601, opened 2019-01-16
Reduce memory usage by changing const strings to string literals.

hexfet:
For the t8sg_v2_plus build:
  
Names #defined as string literals
ROM: 0x08003000 - 0x0803b2c4 = 224.69kB
RAM: 0x20000000 - 0x20008df8 =  35.49kB

Names as char array (master at e33da01e)
ROM: 0x08003000 - 0x0803b34c = 224.82kB
RAM: 0x20000000 - 0x20008e00 =  35.50kB

### DIFF
--- a/src/config/model.c
+++ b/src/config/model.c
@@ -41,44 +41,40 @@ const char * const RADIO_TX_POWER_VAL[TXPOWER_LAST] =
 #define WRITE_FULL_MODEL 0
 static u8 auto_map;
 
-const char MODEL_NAME[] = "name";
-const char MODEL_ICON[] = "icon";
-const char MODEL_TYPE[] = "type";
-const char MODEL_TEMPLATE[] = "template";
-const char MODEL_AUTOMAP[] = "automap";
-const char MODEL_MIXERMODE[] = "mixermode";
+#define   MODEL_AUTOMAP     "automap"
+#define   MODEL_MIXERMODE   "mixermode"
 
-/* Section: Radio */
-static const char SECTION_RADIO[]   = "radio";
+/*        Section:   Radio           */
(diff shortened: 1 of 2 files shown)


PhracturedBlue:
interesting.  I still don't like this approach, but clearly there is some benefit to it.  I'll be interested in whether we see the same reductions when combined with a newer compiler

hexfet:
Results with latest compiler below.  Seems like the only advantage is size.  Thought it would make the file look cleaner but it doesn't.  Perhaps a typedef would be good for that.  Don't mind to close this and just keep in mind if bytes get short again.

With
arm-none-eabi-gcc (GNU Tools for Arm Embedded Processors 8-2018-q4-major) 8.2.1 20181213 (release)
[gcc-8-branch revision 267074]

Names #defined as string literals (const_to_define)
ROM: 0x08003000 - 0x0803ae2c = 223.54kB
RAM: 0x20000000 - 0x20008e00 =  35.50kB

Names as char array (master at e33da01)
ROM: 0x08003000 - 0x0803aea4 = 223.66kB
RAM: 0x20000000 - 0x20008e00 =  35.50kB

hexfet:
lto8.2 + const_to_define
ROM: 0x08003000 - 0x08039830 = 218.05kB
RAM: 0x20000000 - 0x20008d68 =  35.35kB

Please consider an environment variable to disable lto when running make.  It adds significantly to the build time so it will be good to have an option to build without it during development.

PhracturedBlue:
not really a discussion for this thread, but that will be challenging for at least the 7e, since we'll likely take advantage of the extra space it provides...

howard0su:
sorry, 7e LTO build is not done yet. I didn't get it working....

PR #599, opened 2019-01-15
Move ISR code into a seperate file and make sure it exist in main ROM

howard0su:

PR #604, opened 2019-01-16
Enable SBUS and PXX protocols in modular builds

hexfet:
With the merge of #599 these protocols now work in modular builds.

No telemetry for PXX due to memory size limitation.


hexfet:
No, prefer to keep these protocols separate as they are from different origins, unlike dsm2 and dsmx for example.  Just easier to reason about separately.

I do agree that some code is duplicated and it would be good to consolidate that code, but need a way that also works for modular builds.  Maybe moving to the target/common/devo/protocol directory.

PR #613, opened 2019-01-19
Improve "Tx power" user interface

hexfet:
Changes the Tx power setting in model config to show the power levels specific to the radio in use.  Also expands the radio module concept to external modules (R9M).  Open to ideas for improvement.

One change in the UI is that the Tx power setting may change when scrolling through protocols when crossing the boundary between one radio and the next (the ordering of protocols in protocol.h has protocols grouped by radio).  For example, if WFLY (cyrf radio) is selected with Tx power set to 150mW, changing the protocol to Flysky (a7105) will change the power to 100mW because that is the max for a7105.  Then changing back to WFLY, the Tx power will stay at 100mW because that's the 7th power level for both cyrf and a7105.  To address this the radio and it's power setting could be separated from Model into an associated Radio struct, but that seems like overkill.  It doesn't seem a big issue since once a protocol is chosen it's not often changed.

When reading existing model files a power setting may be encountered that is no longer valid for the radio used by the model's protocol.  For backwards compatibility in these cases the power is set to the maximum power for the model's radio.

The R9M is a bit awkward because the power settings depend on the selected radio mode (FCC or EU). I squeezed both values into the Tx power field.  Still thinking on a nice way to handle this.


hexfet:
After sleeping on it I think it's better to keep the Tx power in the model.  One reason is because the same protocol is often used for multiple models.  The popular tiny whoop indoor quads come with Frsky or Flysky built-in for example.  I have both indoor and outdoor helis on dsmx, and the models have different power settings.  Second reason is it would be too much of a breaking change.  No way to stay backwards compatible, with different power levels per model.

TheRealMoeder:
I agree. What about always setting maximum power when scrolling through protocols for safety and consistency reasons?

src/config/model.c
```diff
@@ -672,13 +687,18 @@ int assign_int(void* ptr, const struct struct_map *map, int map_size)
...
+                return 1;
+            }
+            for (i = 0; i < RADIO_TX_POWER_COUNT[m->radio]; i++) {
+                if (MATCH_VALUE(RADIO_TX_POWER_VAL[m->radio][i])) {
                     m->tx_power = i;
                     return 1;
                 }
             }
             printf("Unknown Tx power: %s\n", value);
+            m->tx_power = RADIO_TX_POWER_COUNT[m->radio]-1; // default to radio maximum
```
PhracturedBlue: Lint Error:
```
701:  At least two spaces is best between code and comments  
```

src/protocol/protocol.c
```diff
@@ -91,6 +91,11 @@ const char * PROTOCOL_GetName(u16 idx)
         return Protocols[idx].name;
 }
 
+enum Radio PROTOCOL_GetRadio(u16 idx)
+{
```
PhracturedBlue: Lint Error:
```
95:  { should almost always be at the end of the previous line  
```

src/config/model.c
PhracturedBlue: Lint Error:
```
701:  At least two spaces is best between code and comments  
```

src/protocol/protocol.c
PhracturedBlue: Lint Error:
```
95:  { should almost always be at the end of the previous line  
```

src/config/model.c
hexfet: There's other single space instances in the same file.  Are only changes going through lint?  And only flagged instances should be changed?

src/protocol/protocol.c
hexfet: Agree, but was following style of other functions in that file.  Same questions as above.

hexfet:
Good suggestion.  Updated to set power to radio maximum when protocol is changed.

src/config/model.c
```diff
@@ -29,8 +29,22 @@ struct Model Model;
 static u32 crc32;
 
 const char * const MODEL_TYPE_VAL[MODELTYPE_LAST] = { "heli", "plane", "multi" };
-const char * const RADIO_TX_POWER_VAL[TXPOWER_LAST] =
-     { "100uW", "300uW", "1mW", "3mW", "10mW", "30mW", "100mW", "150mW" };
+const char * const RADIO_TX_POWER_VAL[TX_MODULE_LAST][TXPOWER_LAST] = {
+     { "100uW", "300uW", "1mW", "3mW", "10mW", "30mW", "100mW", "150mW" },  //   CYRF6936,
+     { "100uW", "300uW", "1mW", "3mW", "10mW", "30mW", "100mW" },           //   A7105,
```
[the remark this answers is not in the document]
howard0su: it is preferred to store a number instead of a string.

src/config/model.c
```diff
@@ -29,8 +29,41 @@ struct Model Model;
...
+
+    switch (radio) {
+    case CYRF6936:
+    case CC2500:
+    case A7105:
+    case MULTIMOD:
+        sprintf(tempstring, "%d%cW", std_powers[power], (power < 2)?'u':'m');
+        break;
+    case NRF24L01:
+        sprintf(tempstring, "%dmW", nrf_powers[power]);
```
PhracturedBlue: Lint Error:
```
55:  Never use sprintf. Use snprintf instead.  
58:  Never use sprintf. Use snprintf instead.  
```

src/config/model.c
```diff
@@ -29,8 +29,41 @@ struct Model Model;
...
+
+    if (radio >= TX_MODULE_LAST) return NULL;
+    if (power >= RADIO_TX_POWER_COUNT[radio]) power = RADIO_TX_POWER_COUNT[radio]-1;
+
+    switch (radio) {
+    case CYRF6936:
+    case CC2500:
+    case A7105:
+    case MULTIMOD:
+        snprintf(tempstring, TEMPSTRINGLENGTH, "%d%cW", std_powers[power], (power < 2)?'u':'m');
```
PhracturedBlue: This is pretty ugly.  Is it really less rom to do this than to just index the list of strings for this case?

hexfet: Below are numbers from devo10 build.  The "reduced strings" stores each string only once.  I'm surprised snprintf doesn't take more rom.

```
Full strings
ROM: 0x08004000 - 0x0803bae4 = 222.72kB
RAM: 0x20000000 - 0x20008d8c =  35.39kB

Reduced strings
ROM: 0x08004000 - 0x0803ba84 = 222.63kB
RAM: 0x20000000 - 0x20008d8c =  35.39kB

Numbers (sprintf)
ROM: 0x08004000 - 0x0803ba74 = 222.61kB
RAM: 0x20000000 - 0x20008d8c =  35.39kB

Numbers (snprintf)
ROM: 0x08004000 - 0x0803ba80 = 222.62kB
RAM: 0x20000000 - 0x20008d8c =  35.39kB
```

PhracturedBlue: I'm willing to spend 16 bytes for the nicer code.  Lets do the reduced strings version

TheRealMoeder:
Can you do a final rebase before merge? I‘m interested in the ROM cost  now

src/config/model.c
hexfet: Reverted to reduced strings code.

hexfet:
Rebased.

hexfet:
I agree and considered it but didn't make the change for a few reasons:
- Didn't want to mix GUI changes with this PR
- Other fields on the page should probably be moved also, like PPM In.  I'd put protocol right under name.
- Most folks are using the nightly as they would an official release and this would be a surprise UI change. After next release would be a good time to revisit.
- Don't think it's a big issue for regular users - protocol for a model is not often changed.

PR #619, opened 2019-01-20
Per datasheet to adjust delay for tw8816

howard0su:


howard0su:
Validate on my own Tx

src/target/devof12e/tw8816.c
```diff
@@ -299,9 +299,9 @@ static void TW8816_Init_Ports()
 static void TW8816_Reset()
 {
```
PhracturedBlue: Lint Error:
```
300:  { should almost always be at the end of the previous line  
```

howard0su: lots of work :)

src/target/devof12e/tw8816.c
```diff
@@ -16,23 +16,22 @@
...
+#include <libopencm3/stm32/i2c.h>
 #include <libopencm3/stm32/rcc.h>
-#include "common.h"
 #include "lcd.h"
 
-//Set this to the number of trials
+// Set this to the number of trials
 #define DEBUG_SCREEN_ALIGNMENT 0
 #include "tw8816_init_1.6.h"
 //#include "tw8816_init_1.5.h"
```
PhracturedBlue: Lint Error:
```
28:  Should have a space between // and comment  
29:  Should have a space between // and comment  
30:  Should have a space between // and comment  
31:  Should have a space between // and comment  
```

PhracturedBlue:
well...it is really difficult to review this patch with the style fixes applied.
Also, I don't like some of the formatting changes (like moving curly-braces to the same line...This is one of those things that I think is context specific.  I think I want to separate cleanup patches from code-change patches.

PhracturedBlue:
what is the plan for this?
My concern is that there are many variants of the tw8816 (I think we've seen 6 different revisions in use by Walkera).  They all had different quirks, so I'm very reluctant to make a change that the data-sheet says is ok without having confidence it is ok on multiple actual transmitters.  On the other hand it is almost 0.5sec of wait that you can remove which is quite a lot.  I need to check my F12E and see what revision of the TW8816 it has.

howard0su:
i think I should revert the change on msleep. 

this change was intent to remove Delay calls. I am thinking to replace Delay with usleep(1) which is almost same as the amount of time Delay actual cause.

PR #622, opened 2019-01-20
Revert langv2

howard0su:
Leave the code as it is but disable it from build. The new format prevent user to modify the translation file directly. Introduce the new format when we really need it later.


howard0su:
> you also need to change_extract_strings.pl to set '$format' to default to "v1"

good catch. fixed.

PR #629, opened 2019-01-21
Enable nightlies to upload/download from transifex and push to github.

PhracturedBlue:
Similar functionality to deviation-manual.  When a nightly build is triggered:
* rebuild strings
* upload strings to transifex (only if there are non-comment changes)
* download strings from transifex
* remove any comment-only changes
* push changes to github
* don't re-trigger build on pushed translation changes


PhracturedBlue:
Note that while nightlies can be triggered via the API via the travis webpage (by developers), the cron is not yet enabled.  I still need to write some code to not update the nightlies on deviationtx.com if there are no changes to be uploaded.  Once that is in place, the cron's can be enabled and I'll disable my nightly manual API trigger

PR #630, opened 2019-01-22
don't use fread to skip

howard0su:


howard0su:
it is odd that f12e increase 120bytes...

howard0su:
because fseek is not used anywhere else on devof12e.

But I still like this change even it increase f12e about 120bytes.

PR #633, opened 2019-01-22
No need install libc-i386

howard0su:
After compiler upgrade, the new GCC is a 64bit application instead of a 32bit like GCC 4.8. Remove the package libc6-i386.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,6 @@ addons:
     update: false
     packages:
       - build-essential
-      - libc6-i386
       - mingw32
       - mingw32-binutils
       - mingw32-runtime


PhracturedBlue:
we should probably try moving to xenial sometime.  It will require moving to mingw-64 (which should still generate 32bit binaries), but I don't have 32 bit machines to test with (in the past we had trouble with 32bit binaries built on linux working in 64-bit windows but not 32-bit windows).

It's also been several years, and I'm not sure if 32-bit windows is still prevalent enough to support.

howard0su:
based on my knowledge, it is complete fine to drop 32bit support. If we want to give user the experience to try, I would suggest to think about compile something for javascript. so we can host the emulator in website. No download needed, no struggle on platform support. and it will be a fun project to work with.

PhracturedBlue:
I tried that once.  I had issues, but I don't recall what they were

PR #637, opened 2019-01-22
move switchassign_page sepecfic variables into pagemem

howard0su:


howard0su:
more cleanup is pending for standard mixer code.

src/pages/common/mixer_page.h
```diff
@@ -4,6 +4,8 @@
 #include "mixer.h"
 #include "gui/gui.h"
 #include "buttons.h"
+#include "mixer_standard.h"
```
PhracturedBlue: What is this for?

howard0su: for SWITCHFUNC_LAST which is used later

PR #638, opened 2019-01-22
Use visibility attributes to export symbols to modules

howard0su:
1. The exported functions to Module have to be explicit marked by MODULE_CALLTYPE.
2. Generate the exported symbols table based on the functions marked in 1. All the global variables are still exported by default.
3. All libc functions are directly linked into module to avoid dirty hack to change the original definition of libc functions. So far all modules are not suffering the size problem because this change.


PhracturedBlue:
Not sure how, but is there a way to auto-generate ProtoStubs() from VISIBLE macros, or to otherwise enforce that every VISIBLE exists in ProtoStubs?

howard0su:
I am also thinking about that. We can explicitly control what functions we can let module to call. as today, that requires some cleanup in the module code. The current modules call lots of different functions.

howard0su:
I found the solution to do that:
1. generate a list of functions to export
grep -R VISIBLE *.h --exclude common.h | awk -F"[: ()]" '{print $4}'
2. get the symbols from map
3. use 'comm' to find the intersect.

howard0su:
@PhracturedBlue you can review the PR by each commit. this PR can further break down to smaller ones.

howard0su:
after the change, all the functions module is trying to call is marked as long_call. We can remove the file level long_call pragma. Need some testing here.

PhracturedBlue:
The modules are growing by a minimum of 20bytes and maximum of ~900 bytes with most in the 40ish range.  I'm not really keep on module size growth, since this has historically been the most difficult part of handling the devo7e.  Just because it fits today doesn't mean we should accept growth unless it is brining real gains.

howard0su:
I believe we remove unnecessary long_call pragma should win those bytes back. Let me run some testings. worse case is hack memcpy etc prototypes.

howard0su:
900 bytes should related to the change to include libc.

howard0su:
20bytes are because:
.init           0x0000000020004e7c        0xc
 .init          0x0000000020004e7c        0xc /home/junsu/gcc-arm-none-eabi-8-2018-q4-major/bin/../lib/gcc/arm-none-eabi/8.2.1/crti.o
                0x0000000020004e7c                _init

.fini           0x0000000020004e88        0xc
 .fini          0x0000000020004e88        0xc /home/junsu/gcc-arm-none-eabi-8-2018-q4-major/bin/../lib/gcc/arm-none-eabi/8.2.1/crti.o
                0x0000000020004e88                _fini

Let me try to remove long_call to see what's the gain.

PhracturedBlue:
Looks good

howard0su:
I repurpose this PR to just generate proto_stubs on the fly. And defer the change to only expose certain functions to a module and remove the needs of long_call.

@PhracturedBlue please review.

howard0su:
@PhracturedBlue Can you check?

howard0su:
Missing a function will cause a error during linking module like today.

The next step is only exposing the functions we explictly allow module to call as the last revision I did.

PR #648, opened 2019-01-23
Add special Devo telemetry format

### LINKED ISSUE #627: GPS Telemetry data is incorrect
here is the user reported issue: 
https://www.deviationtx.com/forum/6-general-discussions/8209-problems-with-devo-f12e-nightly-builds#72446

The issue is reported by F12E. I looked at the code. Looks like GPS coord and GPS time are both handled in the generic way.

hexfet:
Change telemetry protocol option to support both standard (Devo) and QR-X350 telemetry formats.  The X350 sends lat/long as decimal degrees.  Standard format is degrees decimal minutes.

Fixes #627.

### DIFF
--- a/src/protocol/devo_cyrf6936.c
+++ b/src/protocol/devo_cyrf6936.c
@@ -51,7 +51,7 @@
 #define NUM_WAIT_LOOPS (100 / 5) //each loop is ~5us.  Do not wait more than 100us
 
 static const char * const devo_opts[] = {
-  _tr_noop("Telemetry"),  _tr_noop("On"), _tr_noop("Off"), NULL,
+  _tr_noop("Telemetry"), _tr_noop("Std"), _tr_noop("X350"), _tr_noop("Off"), NULL,
   NULL
 };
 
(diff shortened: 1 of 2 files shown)


src/protocol/devo_cyrf6936.c
```diff
@@ -615,7 +630,7 @@ const void *DEVO_Cmds(enum ProtoCmds cmd)
             PROTOCOL_Init(0);  // only 1 prot_ops item, it is to enable/disable telemetry
             break;
         case PROTOCMD_TELEMETRYSTATE:
-            return (void *)(long)(Model.proto_opts[PROTOOPTS_TELEMETRY] == TELEM_ON ? PROTO_TELEM_ON : PROTO_TELEM_OFF);
+            return (void *)(long)(Model.proto_opts[PROTOOPTS_TELEMETRY] != TELEM_OFF ? PROTO_TELEM_ON : PROTO_TELEM_OFF);
```
PhracturedBlue: Lint Error:
```
633:  Use int16/int64/etc, rather than the C type long  
```

PR #650, opened 2019-01-24
Devo prototocol cleanup

howard0su:
handle PROTOCMD_SETOPTIONS in the right way to avoid calling back to PROTOCOL_INIT. This can result a recursive call which we should avoid. And PROTOCOL_INIT is not supposed to expose to protocol anyway.

### DIFF
--- a/src/protocol/devo_cyrf6936.c
+++ b/src/protocol/devo_cyrf6936.c
@@ -569,9 +569,18 @@ static void devo_bind()
     PROTOCOL_SetBindState(0x1388 * 2400 / 1000); //msecs
 }
 
-static void initialize()
+static void devo_start()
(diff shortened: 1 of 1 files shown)


howard0su:
@hexfet can I help me verify on real receiver? I don't have devo reciever.

hexfet:
Yes I will check this evening.

hexfet:
This change works fine.  Tested changing from telemetry on to off and back, though I don't have a telemetry receiver.

While testing found a bug in master.  Navigate to the fixed id page, click done, then ext button no longer works, protocol options not accessible with ENT but clicking bind works.  Bug is in devo10 emulator too.  Starting to look at it.

PR #651, opened 2019-01-24
Remove MODULE_CALLTYPE on callback functions

howard0su:
They are not needed. As GCC document states clearly that all the calls
through function pointer are always long-call.

### DIFF
--- a/src/protocol/bayang_nrf24l01.c
+++ b/src/protocol/bayang_nrf24l01.c
@@ -383,7 +383,7 @@ static void bay_init()
 
 
 
-MODULE_CALLTYPE static u16 bay_callback()
+static u16 bay_callback()
 {
     switch (phase) {
     case Bayang_INIT1:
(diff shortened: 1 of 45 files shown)


howard0su:
Verified this is a noop:
$ diff -r devo7e devo7e.orig/
Binary files devo7e/devo7e.elf and devo7e.orig/devo7e.elf differ
diff -r devo7e/hgstamp devo7e.orig/hgstamp
1c1
< devo7e-v5.0.0-18fc37b
---
> devo7e-v5.0.0-ccd65e5
Binary files devo7e/version.o and devo7e.orig/version.o differ

PhracturedBlue:
This was added in: 3f09a47138c5363f22402f08eaa18140e7c08c57
With the log message:    Warning cleanup

It looks like this was needed with the previous protocol design where we used a case statement to call the protocol functions.  The current function-pointer implementation using *_Cmds() shouldn't need it anymore

PR #652, opened 2019-01-25
Fix modal tracking bug introduced in PR #543

hexfet:
After commit b7e1da19 if the model name or fixed id submenus are used the model config page can no longer be exited with EXT and ENT doesn't work to enter protocol options.  Seems like the modal flag is used to keep track of being in these submenus.


howard0su:
shall we just revert the change? Looks like I didn't fully understand this flag.

hexfet:
The gui code is not my strength either.  Let's see what other's comment.

Even before the changes the behavior is not ideal because when the submenus are exited the "previous" page is restarted without any current cursor location.

PR #653, opened 2019-01-25
Cleanup USB stack to save some ROM

howard0su:
1. Split out USB_Enable from [MSC|HID]_Enable
2. Remove some duplicate code
3. remove the array EpInt_IN/OUT.

### DIFF
--- a/src/misc.c
+++ b/src/misc.c
@@ -183,7 +183,7 @@ void wait_release()
 
 void USB_Connect()
 {
-    USB_Enable(0, 1);
+    MSC_Enable();
     //Disable USB Exit
     while(1) {
         if(PWR_CheckPowerSwitch())
@@ -192,6 +192,6 @@ void USB_Connect()
     wait_release();
     wait_press();
     wait_release();
-    USB_Disable();
+    MSC_Disable();
 }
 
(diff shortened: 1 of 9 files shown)


howard0su:
after couple of try to integrate usb lib of libopencm3, I finally decide to stay with our current implementation.

the libopencm3 lib is 1KB smaller due to lack of MSC bulk transfer support. In the same time, the memory usage is about 4KB larger than current usb stack. I tried couple of methods to reduce it. But turns out without big rewrite, it is hard to reduce the memory usage.

So I decided to go back and optimize our current stack to get some additional ROM & RAM saving.

For your reference (stats on devo7e):
integrating libopencm3 usb + umass +hid
ROM: 0x08003000 - 0x0801f228 = 112.54kB
RAM: 0x20000000 - 0x2000379c =  13.90kB

what we are today:
ROM: 0x08003000 - 0x0801f61c = 113.53kB
RAM: 0x20000000 - 0x20002714 =   9.77kB

PhracturedBlue:
This is a lot of work for 100bytes of ROM.  The only thing I'm wary of is the changes to usb_conf.h and usb_istr.c.  These changes make it harder to diff against the reference implementation which in turn makes it harder for us to pull in a new reference implementation should we ever want to in the future.

PhracturedBlue:
I see from your other PR that you intend to do a lot of work to the USB drivers, so I'm not going to sweat the little things.

howard0su:
Please review again. I keep the USB lib code untouched this time.

PR #654, opened 2019-01-25
Remove chip lib in usb stack

howard0su:
the code change is very small. the only logic change is use gpio_set/clear instead of chip lib. Other changes are all about the typedef stuffs.


src/target/common/devo/msc2/hw_config.h
```diff
@@ -16,31 +16,22 @@
...
 #define BULK_MAX_PACKET_SIZE  0x00000040
 /* Exported macro ------------------------------------------------------------*/
 /* Exported functions ------------------------------------------------------- */
 void Set_System(void);
 void Set_USBClock(void);
 void Enter_LowPowerMode(void);
 void Leave_LowPowerMode(void);
-void USB_Cable_Config (u8 enable);
-void USB_Interrupts_Config(void);
+void USB_Cable_Config (uint8_t enable);
```
PhracturedBlue: Lint Error:
```
33:  Extra space before ( in function call  
```

howard0su:
ccceae6 and 81e8881 can be split into another PR.

after cleanup, the code is much easy to organize. I am thinking to split out some umass code into a module to save the rom for devo7e. since when having USB, we anyway disable the protocol module from running. 

I expect we can save 3KB ROM. then even without LTO, i have enough ROM to put devo7e localization support. (of course, some pending changes to fix up UI is still required,) @PhracturedBlue

PhracturedBlue:
I don't see any major issues.  The patches will need to be broken down a bit more for inclusion.  I'd like to see the overall gain materialize before committing these changes because this basically diverges us from any existing USB code-base.  That makes it harder to support if we run into issues, so I'd like to see the gains show it to be worth the effort.

howard0su:
sure. I will keep your concern in mind and reduce the diff with the upstream.

PR #660, opened 2019-01-30
Add new font for devof12e.  Fix centering of loaded font on display.

PhracturedBlue:
This should resolve the cyrillic font issue on the devof12e that has been discussed in #607.
Adds new 18bold font with similar type-setting to the internal font.
Fixes the vertical offset of loaded font to match internal font.


src/target/devof12e/lcd.c
```diff
@@ -230,7 +230,7 @@ u16 load_char_font(u32 c)
         const u8 *data = offset++;
         u8 bit = 0;
         // Data is right aligned, drawn top to bottom
-        for (int y = 3; y < height + 3; ++y)
+        for (int y = (CHAR_HEIGHT - height) >= 0 ? (CHAR_HEIGHT - height) : 0; y < CHAR_HEIGHT; y++)
```
howard0su: the logic is not correct, this is my proposal:

    // caculate the top most
    int start, end;

    if (height >= CHAR_HEIGHT) {
         start = 0;
         end = CHAR_HEIGHT;
    } else {
         start = (CHAR_HEIGHT - height) / 2;
         end = start + height;
    }

    // convert font to tw8816 format
    for (int x = 0; x < width; x++)
    {
        const u8 *data = offset++;
        u8 bit = 0;
        // Data is right aligned, drawn top to bottom
        for (int y = start; y < end; ++y)

PhracturedBlue: I'm not sure that the alignment should be centered.  ideally we'd center on the baseline (2px from the bottom on the tw8816). However, the bold18 font has an ascent of 4, so you can't actually center on it (the bold18 font should also have height==CHAR_HEIGHT so it wouldn't matter).  I don't think the ascent info is stored in the fon file though, so I'm not sure we can do any better than the heuristic you proposed.

vladislavy: I don't see great sense to vertically center fonts relatively to 18, it doesn't look good at menus. For me will be better to align fonts to bottom edge.

howard0su: anyway, y < CHAR_HEIGHT may lead to access undefined memory if height < CHAR_HEIGHT.

PR #663, opened 2019-01-31
Use libopencm3 gpio lib for usb stack

howard0su:
Also make sure we call PowerOff when shutdown USB to clean up usb registers.

### DIFF
--- a/src/target/common/devo/msc2/usb_devo8.c
+++ b/src/target/common/devo/msc2/usb_devo8.c
@@ -95,7 +95,7 @@ void MSC_Disable()
 * Output         : None.
 * Return         : None.
 *******************************************************************************/
-void USB_Cable_Config (u8 NewState)
+void USB_Cable_Config(u8 NewState)
 {
   if (NewState != DISABLE)
   {
--- a/src/target/common/devo/msc2/usb_devo8.c
+++ b/src/target/common/devo/msc2/usb_devo8.c
@@ -26,10 +26,14 @@
 #include "usb_core.h"
 #include "usb_bot.h"
 #include "usb_istr.h"
+#include "usb_pwr.h"
 #define LE_WORD(x)		((x)&0xFF),((x)>>8)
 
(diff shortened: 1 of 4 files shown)


src/target/common/devo/msc2/usb_devo8.c
```diff
@@ -99,6 +102,24 @@ void USB_HandleISR()
...
 }
 
+/*******************************************************************************
+* Function Name  : USB_Cable_Config.
+* Description    : Software Connection/Disconnection of USB Cable.
+* Input          : NewState: new state.
+* Output         : None.
+* Return         : None.
+*******************************************************************************/
+void USB_Cable_Config (u8 NewState)
```
PhracturedBlue: Lint Error:
```
112:  Extra space before ( in function call  
```

src/target/common/devo/msc2/usb_devo8.c
```diff
@@ -90,7 +93,7 @@ void USB_Enable(unsigned type, unsigned use_interrupt)
 
 void USB_Disable()
 {
-    gpio_set(GPIOB, GPIO10);
+    PowerOff();
```
howard0su: because original we didn't call PowerOff of USB stack, we didn't shutdown USB completely. we only did "USB_Cable_Config" in the following PowerOff function. that's why we endup increase by 4 bytes in ROM. 

  /* disable all interrupts and force USB reset */
  _SetCNTR(CNTR_FRES);
  /* clear interrupt status register */
  _SetISTR(0);
  /* Disable the Pull-Up*/
  USB_Cable_Config(DISABLE);
  /* switch-off device */
  _SetCNTR(CNTR_FRES + CNTR_PDWN);
  /* sw variables reset */
  /* ... */

howard0su:
please review. @PhracturedBlue

howard0su:
not needed after we switch to libopencm3 usb lib.

PR #665, opened 2019-02-01
[WIP] Reduce voice RAM needs

TheRealMoeder:


src/extended_audio.c
```diff
@@ -97,22 +98,21 @@ u16 AUDIO_CalculateChecksum(u8 *buffer) {
...
 
 #ifdef EMULATOR     // On emulators call mpg123 to play mp3s
     char cmd[70];
     u16 vol_val = Transmitter.audio_vol * 32786/10;
 #ifdef _WIN32
-    sprintf(cmd, "start /B ..\\..\\mpg123 -f %d -q ..\\..\\mp3\\%04d*.mp3 > nul 2>&1", vol_val, voice_map[music].id);
+    sprintf(cmd, "start /B ..\\..\\mpg123 -f %d -q ..\\..\\mp3\\%04d*.mp3 > nul 2>&1", vol_val, id);
 #else
-    sprintf(cmd, "mpg123 -f %d -q ../../mp3/%04d*.mp3 > /dev/null 2>&1 &", vol_val, voice_map[music].id);
+    sprintf(cmd, "mpg123 -f %d -q ../../mp3/%04d*.mp3 > /dev/null 2>&1 &", vol_val, id);
```
PhracturedBlue: Lint Error:
```
113:  Never use sprintf. Use snprintf instead.  
115:  Never use sprintf. Use snprintf instead.  
```

src/config/model.c
```diff
@@ -1066,7 +1067,8 @@ int assign_int(void* ptr, const struct struct_map *map, int map_size)
 
     if (MATCH_SECTION(SECTION_VOICE)) {
         u16 val = atoi(value);
-        if(val>MAX_VOICEMAP_ENTRIES-1 || voice_map[val].duration == 0 || val < CUSTOM_ALARM_ID) {
+        CONFIG_VoiceParse(val);
+        if(current_voice_mapping.duration == 0 || val < CUSTOM_ALARM_ID) {
```
PhracturedBlue: Lint Error:
```
1071:  Missing space before ( in if(  
```

src/music.h
```diff
@@ -113,7 +113,12 @@ struct VoiceMap {
     u16 duration;
 };
 
-extern struct VoiceMap voice_map[MAX_VOICEMAP_ENTRIES];
+struct AudioQueue {
+    u16 id;
+    u16 duration;
+};
+
+//extern struct VoiceMap voice_map[MAX_VOICEMAP_ENTRIES];
```
```
121:  Should have a space between // and comment  
```

src/timer.c
```diff
@@ -222,12 +221,12 @@ void TIMER_Update()
...
                             * Transmitter.countdown_timer_settings.timeup_interval;
 #if HAS_EXTENDED_AUDIO
-                    warn_time += voice_map[MUSIC_ALARM1 + i].duration;
+                    warn_time += Model.timer[MUSIC_ALARM1 + i].duration;
 #endif
                     if (timer_val[i] > warn_time && (timer_val[i] - delta) <= warn_time) {
 #if HAS_EXTENDED_AUDIO
                         MUSIC_PlayValue(MUSIC_GetTimerAlarm(MUSIC_ALARM1 + i),
-                            (timer_val[i]-voice_map[MUSIC_GetTimerAlarm(MUSIC_ALARM1 + i)].duration)/-1000+1,VOICE_UNIT_TIME,0);
+                            (timer_val[i]-Model.timer[MUSIC_ALARM1 + i].duration)/-1000+1,VOICE_UNIT_TIME,0);
```
PhracturedBlue: Lint Error:
```
229:  Missing space after ,  
```

src/config/voice.c
```diff
@@ -47,31 +46,20 @@ static int ini_handler(void* user, const char* section, const char* name, const
...
-                    voice_map[i].duration = duration;
-                    voice_map[i].id = i;
-                    return 1;
-                }
-            }
-        }
         if (MATCH_SECTION(SECTION_VOICE_CUSTOM)) {
-            voice_map[voice_map_entries].duration = duration;
-            voice_map[voice_map_entries].id = id;
+            //Initiai count of custom voicemap entries
```
PhracturedBlue: Lint Error:
```
59:  Should have a space between // and comment  
```

src/config/voice.c
```diff
@@ -98,11 +86,11 @@ const char* CONFIG_VoiceParse(unsigned id)
         voice_map_entries = CUSTOM_ALARM_ID;  // Reserve space in map for global alerts
         if (CONFIG_IniParse(filename, ini_handler, &id)) {
             printf("Failed to parse voice.ini\n");
-            Transmitter.audio_player = AUDIO_NONE;  // disable external voice output
+            //Transmitter.audio_player = AUDIO_NONE;  // disable external voice output
```
PhracturedBlue: Lint Error:
```
89:  Should have a space between // and comment  
```

TheRealMoeder:
Making Progress, but still pretty heavy on the ROM:

devo7e without standard GUI:
ROM 104.92 kB RAM 9.75 kB

devo7e w/out standard GUI, HAS_EXTENDED_AUDIO
ROM 107.96 kB RAM 10.05 kB

devo7e w/out standard GUI, HAS_EXTENDED_AUDIO && HAS_MUSIC_CONFIG
ROM 108.64 kB RAM 10.05 kB

src/config/voice.c
```diff
@@ -32,7 +32,9 @@ const char SECTION_VOICE_CUSTOM[] = "custom";
 
 static int ini_handler(void* user, const char* section, const char* name, const char* value)
 {
-    // ini_handler returns a requested mp3 label passed at *user to tempstring
+#if !HAS_MUSIC_CONFIG
```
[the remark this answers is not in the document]
TheRealMoeder: Yes. Otherwise, if we build with
HAS_EXTENDED_AUDIO 1
HAS_MUSIC_CONFIG 0
compiler will error with unused parameter.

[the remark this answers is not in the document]
howard0su: i mean you can always include this line not matter what macro is defined.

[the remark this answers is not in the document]
TheRealMoeder: Oh...never mind ;)

src/telemetry.c
```diff
@@ -429,19 +431,19 @@ void TELEMETRY_Alarm()
...
                 case TELEM_DSM_GFORCE_YMAX:
                 case TELEM_DSM_GFORCE_ZMAX:
-                case TELEM_DSM_GFORCE_ZMIN: MUSIC_PlayValue(telem_music, telem_value,VOICE_UNIT_GFORCE,2); break;
+                case TELEM_DSM_GFORCE_ZMIN: telem_unit = VOICE_UNIT_GFORCE; telem_prec = 2; break;
 #if HAS_EXTENDED_TELEMETRY
-                case TELEM_DSM_FLOG_RSSI_DBM: MUSIC_PlayValue(telem_music, telem_value,VOICE_UNIT_DB,0); break;
+                case TELEM_DSM_FLOG_RSSI_DBM: telem_unit = VOICE_UNIT_DB; break;
 #endif
-                default: MUSIC_PlayValue(telem_music, telem_value,VOICE_UNIT_NONE,0);
+                //default: telem_unit = VOICE_UNIT_NONE,0);
```
PhracturedBlue: Lint Error:
```
446:  Should have a space between // and comment  
```

src/telemetry.c
```diff
@@ -460,37 +462,36 @@ void TELEMETRY_Alarm()
...
                 case TELEM_FRSKY_VOLT1:
-                case TELEM_FRSKY_VOLT2: MUSIC_PlayValue(telem_music, telem_value,VOICE_UNIT_VOLT,2); break;
+                case TELEM_FRSKY_VOLT2: telem_unit = VOICE_UNIT_VOLT; telem_prec = 2; break;
 #if HAS_EXTENDED_TELEMETRY
                 case TELEM_FRSKY_TEMP1:
-                case TELEM_FRSKY_TEMP2: MUSIC_PlayValue(telem_music, telem_value-20,VOICE_UNIT_TEMP,0); break;
-                case TELEM_FRSKY_RPM: MUSIC_PlayValue(telem_music, telem_value,VOICE_UNIT_RPM,0); break;
-                case TELEM_FRSKY_CURRENT: MUSIC_PlayValue(telem_music, telem_value,VOICE_UNIT_AMPS,2); break;
-                case TELEM_FRSKY_ALTITUDE: MUSIC_PlayValue(telem_music, telem_value,VOICE_UNIT_ALTITUDE,2); break;
+                case TELEM_FRSKY_TEMP2: telem_value -=20; telem_unit = VOICE_UNIT_TEMP; break;
```
PhracturedBlue: Lint Error:
```
468:  Missing spaces around =  
```

src/telemetry.c
```diff
@@ -460,37 +462,36 @@ void TELEMETRY_Alarm()
...
+                case TELEM_FRSKY_RPM: telem_unit = VOICE_UNIT_RPM; break;
+                case TELEM_FRSKY_CURRENT: telem_unit = VOICE_UNIT_AMPS; telem_prec = 2; break;
+                case TELEM_FRSKY_ALTITUDE: telem_unit = VOICE_UNIT_ALTITUDE; telem_prec = 2; break;
 #endif
                 case TELEM_FRSKY_LRSSI:
-                case TELEM_FRSKY_RSSI: MUSIC_PlayValue(telem_music, telem_value,VOICE_UNIT_DB,0); break;
-                default: MUSIC_PlayValue(telem_music, telem_value,VOICE_UNIT_NONE,0);
+                case TELEM_FRSKY_RSSI: telem_unit = VOICE_UNIT_DB; break;
+                //default: telem_unit = VOICE_UNIT_NONE,0);
```
PhracturedBlue: Lint Error:
```
475:  Should have a space between // and comment  
```

src/telemetry.c
```diff
@@ -460,37 +462,36 @@ void TELEMETRY_Alarm()
...
+                case TELEM_CRSF_BATT_CURRENT: telem_unit = VOICE_UNIT_AMPS; telem_prec = 2; break;
+                case TELEM_CRSF_GPS_ALTITUDE: telem_unit = VOICE_UNIT_ALTITUDE; telem_prec = 2; break;
                 case TELEM_CRSF_TX_SNR:
                 case TELEM_CRSF_TX_RSSI:
                 case TELEM_CRSF_RX_SNR:
                 case TELEM_CRSF_RX_RSSI1:
-                case TELEM_CRSF_RX_RSSI2: MUSIC_PlayValue(telem_music, telem_value,VOICE_UNIT_DB,0); break;
-                default: MUSIC_PlayValue(telem_music, telem_value,VOICE_UNIT_NONE,0);
+                case TELEM_CRSF_RX_RSSI2: telem_unit = VOICE_UNIT_DB; break;
+                //default: telem_unit = VOICE_UNIT_NONE,0);
```
PhracturedBlue: Lint Error:
```
490:  Should have a space between // and comment  
```

howard0su:
which function?

TheRealMoeder:
Values below are the "bad" functions in devo7e with and without voice features. I guess MUSIC_PlayValue, TIMER_Update and TELEMETRY_Alarm are excused, since they add some new functionality. I was surprised at the cost for the two new muxes in MIXER_ApplyMixer.

>       2080 CONFIG_WriteModel
>        596 CONFIG_WriteTx
>       2272 CONFIG_WriteModel
>        612 CONFIG_WriteTx

>        252 MIXER_ApplyMixer
>        392 MIXER_ApplyMixer

>        396 MUSIC_PlayValue

>        340 TELEMETRY_Alarm
>        596 TELEMETRY_Alarm

>        472 TIMER_Update
>        644 TIMER_Update

>        280 ini_handler:hardware
>       2492 ini_handler:model
>        172 ini_handler:music
>        712 ini_handler:tx
>        336 ini_handler:hardware
>       2724 ini_handler:model
>        248 ini_handler:music
>        740 ini_handler:tx

src/extended_audio.c
```diff
@@ -68,51 +69,48 @@ void AUDIO_Init() {
...
 // Send a string  to the Audio device.
-void
-AUDIO_Print(char *string) {
+static void AUDIO_Print(char *string) {
   AUDIO_Send((u8 *)string, strlen(string));
 }
 #endif // EMULATOR
 
-void u16ToArray(u16 value, u8 *array){
+static void u16ToArray(u16 value, u8 *array){
```
PhracturedBlue: Lint Error:
```
85:  Missing space before {  
```

src/music.h
```diff
@@ -29,7 +29,7 @@ enum Music {
 };
 
 #if HAS_EXTENDED_AUDIO
-#define MAX_VOICEMAP_ENTRIES 400 // arbitraty chosen
+#define MAX_VOICEMAP_ENTRIES 999 // arbitraty chosen
```
```
32:  At least two spaces is best between code and comments  
```

src/extended_audio.c
```diff
@@ -171,57 +169,61 @@ void AUDIO_SetVolume() {
...
     }
 
     return 1;
 }
 
 int AUDIO_AddQueue(u16 music) {
-    if (num_audio == AUDIO_QUEUE_LENGTH) {
-        printf("Voice: Queue full, cannot add new mp3 #%d\n",music);
+    if (!AUDIO_VoiceAvailable() || num_audio == AUDIO_QUEUE_LENGTH) {
+        printf("Voice: queue full or voice not available, cannot add new mp3 #%d\n",music);
```
PhracturedBlue: Lint Error:
```
215:  Missing space after ,  
```

src/music.c
```diff
@@ -316,31 +314,31 @@ void MUSIC_PlayValue(u16 music, s32 value, u8 unit, u8 prec)
...
-        if(value > 100) {
+        if (value > 100) {
             digits[digit_count++] = value % 100;
             value /= 100;
             digits[digit_count++] = value + 99;
-            if (thousands){
-              digits[digit_count++] = 109; // MP3 for "thousands"
-              digits[digit_count++] = thousands;
+            if (thousands) {
+                digits[digit_count++] = 109; // MP3 for "thousands"
```
PhracturedBlue: Lint Error:
```
331:  At least two spaces is best between code and comments  
```

src/music.c
```diff
@@ -316,31 +308,31 @@ void MUSIC_PlayValue(u16 music, s32 value, u8 unit, u8 prec)
...
```
PhracturedBlue: Lint Error:
```
325:  At least two spaces is best between code and comments  
```

src/config/model.c
```diff
@@ -1483,6 +1487,12 @@ static void clear_model(u8 full)
     }
     Model.ppmin_centerpw = 1500;
     Model.ppmin_deltapw = 400;
+#if HAS_EXTENDED_AUDIO
+    for (int i = 0; i < NUM_TIMERS; i++) {
```
PhracturedBlue: can this be replaced with `if (HAS_EXTENDED_AUDIO)`?  I'd like to avoid #if conditions where they aren't strictly needed

src/config/tx.c
```diff
@@ -349,7 +349,7 @@ void CONFIG_LoadTx()
     CONFIG_LoadHardware();
     CONFIG_IniParse("tx.ini", ini_handler, (void *)&Transmitter);
     crc32 = Crc(&Transmitter, sizeof(Transmitter));
-#if HAS_EXTENDED_AUDIO
+#if HAS_MUSIC_CONFIG
     CONFIG_VoiceParse(MAX_VOICEMAP_ENTRIES);
```
PhracturedBlue: Same question about `if (HAS_MUSIC_CONFIG)`

src/config/voice.c
```diff
@@ -47,33 +47,23 @@ static int ini_handler(void* user, const char* section, const char* name, const
         duration = atoi(ptr + 1);
         }
 
-#if HAS_MUSIC_CONFIG
     if ( k && (req_id != MAX_VOICEMAP_ENTRIES) && (req_id == id) ) {
-        strlcpy(tempstring, value, k+1);
-        tempstring[MAX_VOICE_LABEL] = '\0';  // limit label length
+        current_voice_mapping.duration = duration;
+#if HAS_MUSIC_CONFIG
+        strlcpy(tempstring, value, k+1);  // return a requested mp3 label passed at *user to tempstring
```
PhracturedBlue: here too

src/config/voice.c
```diff
@@ -47,33 +47,23 @@ static int ini_handler(void* user, const char* section, const char* name, const
...
+#endif
         return 1;
     }
-#endif
+#if HAS_MUSIC_CONFIG
     if ( req_id == MAX_VOICEMAP_ENTRIES ) {
```
PhracturedBlue: and here

src/timer.c
```diff
@@ -170,6 +171,11 @@ void TIMER_Update()
             }
         }
         if (timer_state[i]) {
+#if HAS_EXTENDED_AUDIO
+            unsigned timer_music = MUSIC_ALARM1 + i;
```
PhracturedBlue: Anothercase where it looks like an `if` might work

TheRealMoeder:
@PhracturedBlue the other cases won't work, because compiler complains about struct or variables not being defined. But if we do define them by excluding them from the pp if, thecompiler complains because they are not used

TheRealMoeder:
Works like a charm on T8SGV2+

src/config/voice.c
```diff
@@ -95,16 +84,15 @@ const char* CONFIG_VoiceParse(unsigned id)
     char filename[] = "media/voice.ini";
     #endif
     if (id == MAX_VOICEMAP_ENTRIES) {  // initial parse of voice.ini
-        voice_map_entries = CUSTOM_ALARM_ID;  // Reserve space in map for global alerts
         if (CONFIG_IniParse(filename, ini_handler, &id)) {
             printf("Failed to parse voice.ini\n");
-            Transmitter.audio_player = AUDIO_NONE;  // disable external voice output
+            // Transmitter.audio_player = AUDIO_NONE;  // disable external voice output
```
howard0su: why comment out this line?

src/music.h
```diff
@@ -113,10 +108,12 @@ struct VoiceMap {
     u16 duration;
 };
 
-extern struct VoiceMap voice_map[MAX_VOICEMAP_ENTRIES];
+struct AudioQueue {
```
howard0su: I think this is a QueueEntry instead of a Queue

src/pages/common/_voiceconfig_page.c
```diff
@@ -48,6 +44,29 @@ static u8 voiceconfig_getsrctype (u8 idx)
     return VOICE_SRC_MIXER;
 }
 
+static struct CustomVoice *voiceconfig_setsrcptr(u8 idx) {
+    switch (voiceconfig_getsrctype(idx)) {
+        case VOICE_SRC_SWITCH:
+            return &Model.voice.switches[idx - VOICE_SRC_SWITCH];
+            break;
```
howard0su: break is not needed.

TheRealMoeder:
@howard0su screenshot test fails:
`target/tx/other/test/gui_helper.c:58: Screenshot Voice_config content mismatch!`
Emulator looks fine to me though...

TheRealMoeder:
OK.. Without any custom voice entries in voice.ini voice will currently be disabled, and thus voiceconfig page will show up different. I guess this behavior is not quite correct, as we could have global alarms defined in voice.ini, but no custom alarms. We then could still have voice working, but voice config menu is not available. Will have to think of a way to check that a valid voice.ini exists, but not quite sure how to do so yet...

src/config/voice.c
```diff
@@ -47,32 +47,24 @@ static int ini_handler(void* user, const char* section, const char* name, const
...
-                if (MATCH_KEY(tempstring)) {
-                    voice_map[i].duration = duration;
-                    voice_map[i].id = i;
-                    return 1;
-                }
+    if (HAS_MUSIC_CONFIG) {
+        if ( req_id == MAX_VOICEMAP_ENTRIES ) {
+            if (MATCH_SECTION(SECTION_VOICE_GLOBAL)) {
+                voice_map_entries = 0;
+            }
```
PhracturedBlue: Lint Error:
```
60:  Line ends in whitespace.  Consider deleting these extra spaces.  
```

src/config/voice.c
```diff
@@ -47,32 +47,24 @@ static int ini_handler(void* user, const char* section, const char* name, const
         duration = atoi(ptr + 1);
         }
 
-#if HAS_MUSIC_CONFIG
     if ( k && (req_id != MAX_VOICEMAP_ENTRIES) && (req_id == id) ) {
-        strlcpy(tempstring, value, k+1);
-        tempstring[MAX_VOICE_LABEL] = '\0';  // limit label length
+        current_voice_mapping.duration = duration;
+        if (HAS_MUSIC_CONFIG)
+            strlcpy(tempstring, value, k+1);  // return a requested mp3 label passed at *user to tempstring
```
howard0su: Can we use tempstr_cpy () here?

src/config/voice.h
```diff
@@ -1 +1 @@
-const char* CONFIG_VoiceParse();
+const char* CONFIG_VoiceParse(unsigned id);
```
howard0su: Add "#pragma once"

src/extended_audio.c
```diff
@@ -25,9 +26,9 @@
 #endif // EMULATOR
 
 #if HAS_EXTENDED_AUDIO
-u16 voice_map_entries;
-struct VoiceMap voice_map[MAX_VOICEMAP_ENTRIES];
-u16 audio_queue[AUDIO_QUEUE_LENGTH];
+s16 voice_map_entries = -1;
```
howard0su: Can we move this to Transimiter structure? It is shared across a couple of files.

src/timer.h
```diff
@@ -24,6 +24,9 @@ struct Timer {
     u8 resetsrc;
     u16 timer;
     u32 val;
+#if HAS_EXTENDED_AUDIO
+    u16 duration;
+#endif
     enum TimerType type;
     u8 padding_1[3];  //Needed to ensure structure is word aligned
```
howard0su: You may need to reduce padding_1 to 1 bytes as you are using additional 2 bytes. In other words, you don't need to add #if-def around this variable as anyway we are padding.

src/extended_audio.c
TheRealMoeder: yes. also renamed to voice_ini_entries and added enum to avoid magic numbers

src/timer.h
TheRealMoeder: thanks!

src/config/voice.h
TheRealMoeder: added include guards, to be consistent with the rest of code

src/config/voice.c
TheRealMoeder: no, we only copy the part of value which contains the label.

src/config/voice.h
```diff
@@ -1 +1,6 @@
-const char* CONFIG_VoiceParse();
+#ifndef _VOICE_H_
+#define _VOICE_H_
+
+const char* CONFIG_VoiceParse(unsigned id);
+
+#endif
```
PhracturedBlue: Lint Error:
```
6:  Could not find a newline character at the end of the file.  
```

src/extended_audio.c
```diff
@@ -49,7 +49,14 @@ void AUDIO_Init() {
         printf("Voice: UART5 already initialized\n");
         return;
     }
+    else
```
PhracturedBlue: Lint Error:
```
52:  If/else bodies with multiple statements require braces  
```

src/extended_audio.c
```diff
@@ -142,6 +146,16 @@ int AUDIO_Play(u16 music) {
 }
 
 void AUDIO_SetVolume() {
+#if defined BUILDTYPE_DEV
+#if HAS_AUDIO_UART
+    if (!Transmitter.audio_uart)
```
PhracturedBlue: Lint Error:
```
151:  If/else bodies with multiple statements require braces  
```

src/extended_audio.c
```diff
@@ -171,57 +185,71 @@ void AUDIO_SetVolume() {
...
+        AUDIO_ResetQueue();
         AUDIO_SetVolume();
     }
 }
 
-int AUDIO_VoiceAvailable() {
+static int AUDIO_VoiceAvailable() {
+#if defined BUILDTYPE_DEV
+#if HAS_AUDIO_UART
+    if (!Transmitter.audio_uart)
```
PhracturedBlue: Lint Error:
```
211:  If/else bodies with multiple statements require braces  
```

somewhatlurker:
Assuming this is all tested and works well, the code looks fine to me....   bear in mind that I'm pretty inexperienced and don't know the project well yet though.

I see that it only deals in external voice IDs now though...  It's probably good, and half solves #999 in the process, but does introduce the small chance for issues loading old models with an abnormal voice.ini. (though it does make it essentially impossible to add support for gaps in numbering)

It would be possible to adopt a system of referencing voice.ini by line number rather than key names, which would behave more like the old system and allow for an almost identical solution to what I already proposed for #999. (models would take longer to load due to repeated config parsing, but it'd require benchmarking to figure out how significant it is)

TheRealMoeder:
I think the code is significantly better with these changes, but far from good. Yes, I tested them thoroughly and didn't find any issues. Using line numbers is an interesting idea, so before introducing this potentially backwards compatibility breaking change, would you have time to use this branch as a base for reworking #999?

somewhatlurker:
yeah, I'll try to have a look into it; or at least see how much work it would be to make behaviour identical to the old system for reimplementation of #1000 at another time.

somewhatlurker:
Here's a commit that seems to make voice loading match the old behaviour: https://github.com/somewhatlurker/deviation/commit/d34f6f4f0fe5c8b4087b4e10c719e4d4e813907d

Tested using a voice.ini containing the following, exhibiting misordered and missing ids:
```
[custom]
200=armed,548
201=disarmed,783
202=throttle hold on,1175
203=throttle hold off,1149
204=low rates,809
211=acro mode,835
;205=medium rates,1018
;206=high rates,783
207=angle mode 835
;saved hold0 alert as stability mode (207)
208=stability mode,1123
209=horizon mode,966
210=rate mode,679
212=flight mode 0,1253
213=flight mode 1,1149
214=flight mode 2,1097
215=flight mode 3,1175
...
```

Under both current master code and my branch (tested in zip_win_emu_t8sg_v2), `HOLD0=207` in a model's voice section refers to "stability mode" (mp3 208).
Note that saving still uses the external IDs and thus the fix from #1000 would need to be implemented to even have a self-consistent system at this stage.

Another issue is that you can't reach the end of the list due to non-contiguous numbering. That could be fixed by either using the line number referencing in the voice config page or recording the maximum ID rather than number of IDs to `voice_ini_entries`. (the former would behave more like the old system)

somewhatlurker:
An update on what I've been doing:

https://github.com/somewhatlurker/deviation/tree/voice_diet2 has been updated to use essentially the same system as #1000 to load both old and new entries correctly, however it uses the "+" prefix to at least maintain some level of support for loading new models in older firmware. The meaning of the prefix should probably be explained in the manual if these changes are used.

It also makes all IDs up to the maximum used one accessible in model voice config. IDs with no voice will have the label "[ID#] (missing)". (it might make sense to have this translatable, but I'm not sure how to do that)

TheRealMoeder:
Your work looks fine. Translation should be triggered with `_tr("string")`. While your submission should work nice, I'm honestly not sure if it is worth keeping backward compatibility here. We never had an official release with voice, it was always a nightly feature (although that is what 99% uses). The Prefix stuff makes the code unnecessarily bloated, and as you suggested the new prefix is not self-explanatory. @goebish opinion?

somewhatlurker:
Yeah, I see what you mean about bloat after checking the full diff for my changes. I suppose cherry picking 5776c82 and ef11399 (and just the voiceconfig_page.c changes from d34f6f4) to only fix actual issues would be fine if code size is a concern. (though retaining guaranteed compatibility would be nice)

I feel you on the features front, but the audio system still seems pretty daunting to me with all the stuff in there together. Obviously an entirely new design for everything would be best, but that's definitely not something I can help with when I don't even understand the full scope of current features well. Would you want to work on the design or is that just a hypothetical scenario right now?

TheRealMoeder:
My time is limited, but the Music/Voice topic is actually not too big to tackle, so probably something we could easily optimize. For now, I think making a PR out of voice_diet2 with your patches sounds like good idea to me, after that we two (and maybe @eried if he has time) can discuss further development of voice/sound support in a separate issue.

PR #670, opened 2019-02-06
Fix index out-of-bounds and null pointer in power settings

hexfet:
These bugs came in with the recent power setting changes.  Found problem in emulator.  Seemed to have no effect in devo10 or t8sg+.

### DIFF
--- a/src/config/model.c
+++ b/src/config/model.c
@@ -45,21 +45,22 @@ const char * radio_tx_power_val(enum Radio radio, enum TxPower power) {
     static const char * const nrf_powers[] = { "1mW", "6mW", "25mW", "100mW"};
     static const char * const r9m_powers[] = { "10/25mW", "100/25mW", "500/500", "Auto/200" };
 
-    if (radio >= TX_MODULE_LAST) return NULL;
-    if (power >= RADIO_TX_POWER_COUNT[radio]) power = RADIO_TX_POWER_COUNT[radio]-1;
 
     switch (radio) {
     case CYRF6936:
     case CC2500:
     case A7105:
     case MULTIMOD:
+        power = RADIO_TX_POWER_COUNT[radio]-1;
         return std_powers[power];
     case NRF24L01:
+        power = RADIO_TX_POWER_COUNT[radio]-1;
         return nrf_powers[power];
     case R9M:
+        power = RADIO_TX_POWER_COUNT[radio]-1;
         return r9m_powers[power];
     default:
-        return NULL;    // never reached, silence warning
+        return std_powers[TXPOWER_LAST-1];
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/config/model.c
```diff
@@ -45,21 +45,22 @@ const char * radio_tx_power_val(enum Radio radio, enum TxPower power) {
...
         return std_powers[power];
     case NRF24L01:
+        power = RADIO_TX_POWER_COUNT[radio]-1;
         return nrf_powers[power];
     case R9M:
+        power = RADIO_TX_POWER_COUNT[radio]-1;
         return r9m_powers[power];
     default:
-        return NULL;    // never reached, silence warning
+        return std_powers[TXPOWER_LAST-1];
```
howard0su: if this is not possible, better we just return 0 instead of default to std_powers.

src/pages/common/_model_page.c
```diff
@@ -250,7 +250,7 @@ static const char *protoselect_cb(guiObject_t *obj, int dir, void *data)
         // Load() the new protocol
         Model.protocol = new_protocol;
         Model.radio = PROTOCOL_GetRadio(new_protocol);
-        Model.tx_power = RADIO_TX_POWER_COUNT[Model.radio]-1;
+        if (Model.radio < TX_MODULE_LAST) Model.tx_power = RADIO_TX_POWER_COUNT[Model.radio]-1;
```
howard0su: curious, why Model.radio has an invalid value? why not fix the place which cause invalid value instead of check here? The check is more or less a hack here.

src/config/model.c
hexfet: Not sure if you mean return "0" or NULL.  Returning NULL is the problem, as CONFIG_WriteModel expects a string to write to the model file.  I didn't want to put exception logic everywhere the function is used to check for a null pointer.  I didn't add a new string to save on space, and in other places 150mW is used as the power when the protocol doesn't have a radio module or the radio doesn't have a power amp.

src/pages/common/_model_page.c
hexfet: I agree, and we had a couple comments on this in the original PR.  Have not yet come up with a cleaner way to handle the relationship between models, protocols, internal radios, and external radios, that doesn't add a lot of code.  I will try to clean this up but am traveling the next 10 days. Would be better to merge as master is writing invalid power values now to the model files if the protocol doesn't have an associated radio.  Doesn't cause a problem because the value is ignored when read.  The emulator crashes when trying to write.

PhracturedBlue:
This broke the ability to change tx power.  It needs to be revisited

PhracturedBlue:
I have reverted it for now

PR #674, opened 2019-02-09
increase the string count by 10

howard0su:


howard0su:
We should think about avoiding this static variable. my change on devo7e removes the needs of 8K string buffer. but we need do something additional to avoid this lookup table.

PR #675, opened 2019-02-10
Move telemetry data into one structure

howard0su:
the change is motivated by configure module change. With the change, the code can be written in data driven way.


howard0su:
> 
> 
> It looks fine, except for the build fail on devof7

It is an interesting error. When we link first time, we find candidate functions to fit the hole. However during the second time link, one function need some glue and cause the function bigger than original by 2 bytes. Then we got the error. this disappear after I rename some variables.

PR #678, opened 2019-02-11
consolidate usleep implementation to stm32 folder

howard0su:
The assembly output is verified. Gcc produce the code one few byte.
before:
00000000 <_usleep>:
   0:   f04f 0118       mov.w   r1, #24
   4:   fb00 f001       mul.w   r0, r0, r1
   8:   e000            b.n     c <_delaycmp>
0000000a <_delayloop>:
   a:   3801            subs    r0, #1
0000000c <_delaycmp>:
   c:   2800            cmp     r0, #0
   e:   d1fc            bne.n   a <_delayloop>
  10:   4770            bx      lr

after
00000000 <_usleep>:
   0:   2318            movs    r3, #24
   2:   fb00 f003       mul.w   r0, r0, r3
   6:   e000            b.n     a <_delaycmp>
00000008 <_delayloop>:
   8:   3801            subs    r0, #1
0000000a <_delaycmp>:
   a:   2800            cmp     r0, #0
   c:   d1fc            bne.n   8 <_delayloop>
   e:   4770            bx      lr


src/target/common/stm32/clock.c
```diff
@@ -226,3 +226,22 @@ u32 RTC_GetValue()
...
+        ("mul %0, %0, %1;"
+         "b.n _delaycmp;"
+         "_delayloop:"
+         "subs %0, %0, #1;"
+         "_delaycmp:;"
+         "cmp %0, #0;"
+         "bne.n _delayloop;"
+         :
+         : "r" (x), "r" (COUNT)
+         );
```
PhracturedBlue: Lint Error:
```
246:  Closing ) should be moved to the previous line  
```

src/target/common/devo/tx_misc.c
```diff
@@ -16,21 +16,4 @@ void TxName(u8 *var, int len)
...
-         "subs r0, r0, #1;"
-         "_delaycmp:;"
-         "cmp r0, #0;"
-         "bne.n _delayloop;"
-         :
-         : "r" (x)
-         : "r1"
-         );
-}
+}
```
PhracturedBlue: Lint Error:
```
19:  Could not find a newline character at the end of the file.  
```

src/target/common/stm32/clock.c
```diff
@@ -226,3 +226,21 @@ u32 RTC_GetValue()
     return value;
 }
 
+void _usleep(u32 x)
+{
+#if FREQ_MHz == 72
+    #define COUNT 24
+#else
+    #error Need define the count for usleep
+#endif
```
PhracturedBlue: How about something like:
```
# if (FREQ_MHZ % 3)
    #error Frequency is not a multiple of 3 - need custom usleep
# else
   #define COUNT (FREQ_MHz / 3)
#endif
```
It probably won't work properly if we ever see an f4, but we can deal with that when the day comes

PR #679, opened 2019-02-11
Always generate debug symbols

howard0su:
This make postmortem of error.txt easier.


howard0su:
the size of firmware gets a little bigger. Not sure if we are sensitive about the size of downloading or not.

howard0su:
I am dealing with some bug reports recently.

PR #684, opened 2019-02-11
Save ROM on 128x64 timer_page.c

TheRealMoeder:


src/pages/128x64x1/timer_page.c
```diff
@@ -43,37 +43,65 @@ static int row_cb(int absrow, int relrow, int y, void *data)
...
+    // Col 1
+    for (int i = 0; i < 4; i++) {
+        switch (i) {
+            case 0:
+                lbl_obj = &gui->name;
+                lbl_str = timer_str_cb; lbl_data = (void *)(long)absrow;
+                break;
+            case 1:
+                lbl_obj = &gui->switchlbl;
+                lbl_str = switch_str_cb; lbl_data = (void *)(long)absrow;
```
PhracturedBlue: Lint Error:
```
57:  Use int16/int64/etc, rather than the C type long  
61:  Use int16/int64/etc, rather than the C type long  
```

TheRealMoeder: ignored

TheRealMoeder:
Don't know if it's really worth the effort in this case...

howard0su:
i feel the new code is not cleaner.

TheRealMoeder:
That was my thought too

howard0su:
not worth for the gain to compromise the code readability.

TheRealMoeder:
Agree.

PR #685, opened 2019-02-12
Only caculate the out channels that are needed

howard0su:
Based on Model.num_channels to only caculate the channels are using.

### DIFF
--- a/src/mixer.c
+++ b/src/mixer.c
@@ -80,7 +80,7 @@ void MIXER_EvalMixers(volatile s32 *raw)
     int i;
     s32 orig_value[NUM_CHANNELS];
     //3rd step: apply mixers
-    for (i = 0; i < NUM_CHANNELS; i++) {
+    for (i = 0; i < Model.num_channels; i++) {
         orig_value[i] = raw[i + NUM_INPUTS + 1];
     }
     for (i = 0; i < NUM_MIXERS; i++) {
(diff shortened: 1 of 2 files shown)


src/mixer.c
```diff
@@ -80,7 +80,7 @@ void MIXER_EvalMixers(volatile s32 *raw)
     int i;
     s32 orig_value[NUM_CHANNELS];
     //3rd step: apply mixers
-    for (i = 0; i < NUM_CHANNELS; i++) {
+    for (i = 0; i < Model.num_channels; i++) {
         orig_value[i] = raw[i + NUM_INPUTS + 1];
     }
     for (i = 0; i < NUM_MIXERS; i++) {
```
PhracturedBlue: I'm pretty sure this won't work with speed-limits or state-machine mixers

src/mixer.c
```diff
@@ -190,7 +190,7 @@ void MIXER_CalcChannels()
     //4th step: apply auto-templates
     s32 cyclic[3];
     MIXER_CreateCyclicOutput(raw, cyclic);
-    for (i = 0; i < NUM_OUT_CHANNELS; i++) {
+    for (i = 0; i < Model.num_channels; i++) {
         switch(Model.templates[i]) {
```
PhracturedBlue: I'm not sure this will work properly with virtual channels

PR #687, opened 2019-02-12
split num_ppmin into num_ppmin_channels and num_ppmin_mode

howard0su:
The RAM saving is only 1 byte, but the code is much more complex. use 2 bytes and cleanup the code.

### DIFF
--- a/src/mixer.c
+++ b/src/mixer.c
@@ -155,7 +155,7 @@ static void MIXER_UpdateRawInputs()
         raw[i] = CHAN_ReadInput(mapped_channel);
     }
     if (PPMin_Mode() == PPM_IN_SOURCE && ppmSync) {
-        for (i = 0; i < (Model.num_ppmin & 0x3f); i++) {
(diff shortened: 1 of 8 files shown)


howard0su:
motivated by the configure modele cleanup

src/mixer.c
```diff
@@ -141,21 +141,22 @@ static void MIXER_UpdateRawInputs()
 {
     int i;
     //1st step: read input data (sticks, switches, etc) and calibrate
-    for (i = 1; i <= NUM_TX_INPUTS; i++) {
-        unsigned mapped_channel = MIXER_MapChannel(i);
-        if (PPMin_Mode() == PPM_IN_TRAIN2 && Model.train_sw && raw[Model.train_sw] > 0) {
+    if (PPMin_Mode() == PPM_IN_TRAIN2 && Model.train_sw && raw[Model.train_sw] > 0) {
```
howard0su: @PhracturedBlue please double check the change in this file. I think your mixer tests cover here. but better you can review this.

PR #688, opened 2019-02-13
Enable standard mixer page tests real

howard0su:
Now all standard page and voice configure page has the real content.


howard0su:
current tests are implemented as an alternative to devo8. support 128gui need some smart way to do so. I don't have a good idea so far. @PhracturedBlue ?

PR #694, opened 2019-02-16
Driver reorg

PhracturedBlue:
Reorganize driver code to make things more modular.  There will be several steps to this process, including moving the transmitters into a different hierarchy, moving the emu and devo code location, replacing ports.h with hardware.h, building hardware defines from yaml,...  This is just the 1st step.  It should have zero impact on the resulting ROM size.

ToDo:
[x] Move display drivers
[x] Move mcu code
[ ] Move backlight code
[x] move filesystem code


src/target/drivers/display/8080/320x240x16_ili9341.h
```diff
@@ -1,24 +1,89 @@
...
+#define ILI9341_2D_COLORSET_REG           0x2D
+#define ILI9341_2E_MEMORYREAD_REG         0x2E
+#define ILI9341_30_PARTIALAREA_REG        0x30
+#define ILI9341_33_VERTSCROLL_REG         0x33
+#define ILI9341_34_TEAREFFECTLINEOFF_REG  0x34
+#define ILI9341_35_TEAREFFECTLINEON_REG   0x35
+#define ILI9341_36_MEMACCESS_REG          0x36
+#define ILI9341_37_VERSCRSRART_REG        0x37
+#define ILI9341_38_IDLEMODEOFF_REG        0x38
+#define ILI9341_39_IDLEMODEON_REG         0x39
```
PhracturedBlue: Lint Error:
```
35:  Line ends in whitespace.  Consider deleting these extra spaces.  
```

src/target/drivers/display/8080/320x240x16_ili9341.h
```diff
@@ -1,24 +1,89 @@
...
+#define ILI9341_3C_WRITEMEMCONTINUE_REG   0x3C
+#define ILI9341_3E_READMEMCONTINUE_REG    0x3E
+#define ILI9341_44_SETTEATSCAN_REG        0x44
+#define ILI9341_45_GETSCANLINE_REG        0x45
+#define ILI9341_51_WRITEBRIGHT_REG        0x51
+#define ILI9341_52_READBRIGHT_REG         0x52
+#define ILI9341_53_WRITECTRL_REG          0x53
+#define ILI9341_54_READCTRL_REG           0x54  
+#define ILI9341_55_WRITECABC_REG          0x55
+#define ILI9341_56_READCABC_REG           0x56
```
PhracturedBlue: Lint Error:
```
44:  Line ends in whitespace.  Consider deleting these extra spaces.  
46:  Line ends in whitespace.  Consider deleting these extra spaces.  
```

howard0su:
how about usb？

src/target/drivers/display/8080/320x240x16_ili9341.h
```diff
@@ -185,3 +250,4 @@ void ili9341_init()
     Delay(5);
     disp_type = &ili9341_type;
 }
+#endif
```
PhracturedBlue: Lint Error:
```
253:  Uncommented text after #endif is non-standard.  Use a comment.  
```

PhracturedBlue:
I'm going to defer the backlight fixes.  They aren't as easy since they are tied to the mcu.

howard0su:
at9 code increase may cause by this;
    if (type == LCDTYPE_ILI9341) {
        ili9341_init();
    } else if (type == LCDTYPE_HX8347) {
        hx8347_init();

at9 doesn't really need hx8347 code. maybe we need introduce a knob to control the which indeed the driver to compile in.

src/target/drivers/display/8080/320x240x16.c
```diff
@@ -35,17 +37,23 @@ void lcd_set_pos(unsigned int x0, unsigned int y0)
...
 {
-    //Read ID register for HX8347 (will be 0x47 if found)
-    LCD_REG = 0x00;
-    u8 data = LCD_DATA;
-    if (data == 0x47) {
-        return LCDTYPE_HX8347;
-    } else {
+    u8 data;
+    if (HAS_LCD_TYPE(LCDTYPE_HX8347)) {
+        //Read ID register for HX8347 (will be 0x47 if found)
```
PhracturedBlue: Lint Error:
```
49:  Should have a space between // and comment  
```

src/target/drivers/display/8080/320x240x16.c
```diff
@@ -35,17 +37,23 @@ void lcd_set_pos(unsigned int x0, unsigned int y0)
     //printf("lcd_set_pos: %d, %d\n", x0, y0);
 }
 
-#define LCDTYPE_HX8347  1
-#define LCDTYPE_ILI9341 2
-#define LCDTYPE_UNKNOWN 0
+#define LCDTYPE_HX8347  0x01
```
howard0su: can we move this define to some .h file as target_defs.h is referencing it.

src/target/drivers/display/8080/320x240x16.c
```diff
@@ -131,10 +138,21 @@ void LCD_Init()
...
+
+#ifdef ILI9341_RESET_PIN
+    /* Reset pin for ILI9341 */
+    gpio_set_mode(ILI9341_RESET_PIN.port, GPIO_MODE_OUTPUT_50_MHZ,
+                  GPIO_CNF_OUTPUT_PUSHPULL, ILI9341_RESET_PIN.pin);
+
+    gpio_clear(ILI9341_RESET_PIN.port, ILI9341_RESET_PIN.pin);
+    _usleep(10);  // must be held low for at least 10us
+    gpio_set(ILI9341_RESET_PIN.port, ILI9341_RESET_PIN.pin);
+    _msleep(120); // must wait 120ms after reset
```
PhracturedBlue: Lint Error:
```
150:  At least two spaces is best between code and comments  
```

src/target/drivers/display/8080/320x240x16.c
PhracturedBlue: there's not really a good place to put it right now.  Putting it in a common file would just be clutter  to any other tx, and the target_defs is loaded for virtually every c file.  We don't have a define to tell us which display is used, used there isn't a way to set these things based on a header.  Maybe when I rework the target_defs.h or build system it will be easier to define a proper way to handle this without needing to have redundant information.  Ideally screen-size should be set be defining the display-driver, but that too isn't possible right now.

PhracturedBlue:
well, strictly speaking we should try to re-init if an init fails.  This pull request should not change any existing logic though, so I'm not making any changes to how the code functions here

PR #705, opened 2019-02-18
T8SG:Don't trigger watchdog during long-press power-down

PhracturedBlue:


src/target/tx/devo/common/power.c
```diff
@@ -58,7 +58,11 @@ void PWR_Shutdown()
     rcc_set_sysclk_source(RCC_CFGR_SW_SYSCLKSEL_HSICLK);
     rcc_wait_for_osc_ready(RCC_HSI);
     gpio_clear(_PWREN_PORT, _PWREN_PIN);
-    while(1) ;
+    while(1) {
```
PhracturedBlue: Lint Error:
```
61:  Missing space before ( in while(  
```

src/target/tx/devo/common/power.c
```diff
@@ -58,7 +58,11 @@ void PWR_Shutdown()
...
+#if defined(HAS_BUTTON_POWER_ON) && HAS_BUTTON_POWER_ON
+        CLOCK_ResetWatchdog;  // If the reset is held too long, the watchdog could kick in
```
howard0su: CLOCK_ResetWatchdog()?

TheRealMoeder:
We should probably be doing the same when playing MUSIC_Play(MUSIC_SHUTDOWN); in main.c (although this didn't make shutdown music work for me...)

PR #712, opened 2019-02-18
Add support for 250mW power level in CRSF telemetry.

### LINKED ISSUE #669: CRSF protocol telemetry reporting for 250mw mode
Not a major issue with crsf but on the telemetry page there is a value which reports the current transmitter power value.  25mw, 100mw, or 250mw on the micro tx I'm using.  25w and 100w report cleanly, but 250w just reports a string of numbers that extend beyond the box.  Thank you for this amazing addition of crsf to deviation.

hexfet:
Fixes #669

### DIFF
--- a/src/protocol/crsf_uart.c
+++ b/src/protocol/crsf_uart.c
@@ -157,8 +157,8 @@ static void processCrossfireTelemetryFrame()
       for (i=1; i <= TELEM_CRSF_TX_SNR; i++) {
         if (getCrossfireTelemetryValue(2+i, &value, 1)) {   // payload starts at third byte of rx packet
           if (i == TELEM_CRSF_TX_POWER) {
-            static const s32 power_values[] = { 0, 10, 25, 100, 500, 1000, 2000 };
-            value = power_values[value];
+            static const s32 power_values[] = { 0, 10, 25, 100, 500, 1000, 2000, 250 };
+            value = power_values[value % (sizeof power_values / sizeof (s32))];
           }
           set_telemetry(i, value);
         }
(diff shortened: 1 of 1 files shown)


src/protocol/crsf_uart.c
```diff
@@ -157,8 +157,8 @@ static void processCrossfireTelemetryFrame()
       for (i=1; i <= TELEM_CRSF_TX_SNR; i++) {
         if (getCrossfireTelemetryValue(2+i, &value, 1)) {   // payload starts at third byte of rx packet
           if (i == TELEM_CRSF_TX_POWER) {
-            static const s32 power_values[] = { 0, 10, 25, 100, 500, 1000, 2000 };
-            value = power_values[value];
+            static const s32 power_values[] = { 0, 10, 25, 100, 500, 1000, 2000, 250 };
+            value = power_values[value % (sizeof power_values / sizeof (s32))];
           }
```
howard0su: why do we need this? This still lead to wrong value in case of value is out of range.

hexfet: The value may not be right, but it does prevent indexing past the end of the defined values so a valid value is always returned.  The telemetry value is eight bits, and the documentation doesn't include 250mW so I wouldn't be surprised to see other values crop up without warning.

The doc reads : `​uint8_t Uplink TX Power ( enum 0mW = 0, 10mW, 25 mW, 100 mW, 500
mW, 1000 mW, 2000mW )`

hexfet: Changed code to not update telemetry unless known enum value received.

PR #719, opened 2019-02-19
Prototype of layout

howard0su:


src/pages/common/layout.h
```diff
@@ -0,0 +1,16 @@
+#ifndef _LAYOUT_H_
+#define _LAYOUT_H_
+
+typedef struct _grid_layout_data {
+	int w, h, x, y;
```
PhracturedBlue: Lint Error:
```
5:  Tab found; better to use spaces  
```

PhracturedBlue:
overall, I like the idea.  I think we'll need non-uniform grids, and I'd like to see how we can use it for varying width displas (like between the devo8 and devo12) or how we can support the x9d 212x64 display (which we could easily model on one of the b&w devo tx in the emulator)  to see how scalable it is.

PR #720, opened 2019-02-20
Fix warnings noticed by clang

howard0su:


src/telemetry/telem_crsf.c
```diff
@@ -89,7 +89,7 @@ const char * _crsf_short_name(char *str, u8 telem)
     case TELEM_CRSF_TX_POWER: strcpy(str, _tr("Power")); break;
     case TELEM_CRSF_RX_QUALITY: strcpy(str, _tr("rxQual")); break;
     case TELEM_CRSF_TX_QUALITY: strcpy(str, _tr("txQual")); break;
-    case TELEM_CRSF_BATT_VOLTAGE: sprintf(str, _tr("Volt")); break;
+    case TELEM_CRSF_BATT_VOLTAGE: strcpy(str, _tr("Volt")); break;
```
PhracturedBlue: Lint Error:
```
92:  Almost always, snprintf is better than strcpy  
```

PhracturedBlue: google is wrong here.  snprintf is the worst option for us.  strlcpy would be better, but unneeded here.  we should disable this rule in the linter

src/pages/320x240x16/range_page.c
```diff
@@ -37,7 +37,7 @@ static const char *startstop_dlg(struct guiObject *gui, const void *data) {
 static void _draw_page(int has_pa) {
     PAGE_ShowHeader(PAGE_GetName(PAGEID_RANGE));
     if (!has_pa) {
-        snprintf(tempstring, sizeof(tempstring), _tr("No range test possible."));
+        snprintf(tempstring, sizeof(tempstring), "%s", _tr("No range test possible."));
         GUI_CreateLabelBox(&gui->info, INFO_X, INFO_Y, 0, 0, &DEFAULT_FONT,
```
PhracturedBlue: I think strlcpy would be better

src/pages/common/_debuglog_page.c
```diff
@@ -28,13 +28,13 @@ static u32 changed = 0;
...
-    unsigned idx = (long)data;
+    unsigned idx = (unsigned)data;
     if (idx == 255) {
         return "";
     }
     const char *dptr = (char *)&logstr[line_pos[idx]];
     const char *w = (char *)wptr;
-    sprintf(tempstring, "%d:", (long)data);
+    sprintf(tempstring, "%d:", (unsigned)data);
     char *ptr = tempstring + strlen(tempstring);
```
PhracturedBlue: this should probably be '%u' if using an unsigned

src/pages/common/_debuglog_page.c
```diff
@@ -28,13 +28,13 @@ static u32 changed = 0;
 static const char *str_cb(guiObject_t *obj, const void *data)
 {
     (void)obj;
-    unsigned idx = (long)data;
+    unsigned idx = (unsigned)data;
     if (idx == 255) {
```
PhracturedBlue: This should be uintptr_t.  you shouldn't cast a pointer to an int.

src/pages/common/_model_loadsave.c
```diff
@@ -30,7 +30,7 @@ static int ini_handle_icon(void* user, const char* section, const char* name, co
 
 static int ini_handle_name(void* user, const char* section, const char* name, const char* value)
 {
-    long idx = (long)user;
+    int idx = (int)user;
     if(section[0] == '\0' && (strcasecmp(name, MODEL_NAME) == 0 || strcasecmp(name, MODEL_TEMPLATE) == 0)) {
```
PhracturedBlue: what is wrong with this?  sizeof(void*) == sizeof(long) on all platforms

src/pages/common/_timer_page.c
```diff
@@ -189,7 +189,7 @@ const char *show_timerperm_cb(guiObject_t *obj, const void *data)
   (void)obj;
   u8 idx = (long)data;
   if (idx & 0x80)
-      snprintf(tempstring, sizeof(tempstring), _tr("Set to"));
+      snprintf(tempstring, sizeof(tempstring), "%s", _tr("Set to"));
   else
```
PhracturedBlue: strlcpy again

src/protocol/cflie_nrf24l01.c
```diff
@@ -343,7 +343,7 @@ static void send_crtp_rpyt_packet()
         float pitch;
         float yaw;
         uint16_t thrust;
-    }__attribute__((packed)) cpkt;
+    }cpkt;
```
PhracturedBlue: why did you remove the packed attribute.  I don't know this protocol, but it is likely the packet is strictly defined as far as its structure goes, and the compiler can't insert any padding)

src/protocol/inav_nrf24l01.c
```diff
@@ -85,7 +85,7 @@
 
 #ifdef EMULATOR
 #define USE_FIXED_MFGID
-#define BIND_COUNT 4.360   // 6 seconds
+#define BIND_COUNT 4      // 6 seconds
 #define FIRST_PACKET_DELAY  1
```
PhracturedBlue: this just looks weird.  was the original a typo?

src/protocol/symax_nrf24l01.c
```diff
@@ -41,7 +41,7 @@
 
 #ifdef EMULATOR
 #define USE_FIXED_MFGID
-#define BIND_COUNT 4.360   // 6 seconds
+#define BIND_COUNT 4      // 6 seconds
 #define FIRST_PACKET_DELAY  1
```
PhracturedBlue: same question as above

src/pages/common/_model_loadsave.c
howard0su: long should use %ld to printf. I am not sure if we want to do that in the following line. and idx to me int is large enough to hold idx.

src/protocol/cflie_nrf24l01.c
howard0su: there is risk for having packed structure on ARM platform as not ARM instruction support unaligned access. this get warned in the below code which can result unaligned access.

remove packed is not a right solution, i agree. let me tune the code a bit to manually pack the code to avoid this code accidently get broken by some compiler udpate.

src/protocol/inav_nrf24l01.c
howard0su: I am not quite sure. @goebish can u help understand why this is a float which is expect to be an integer.

src/timer.c
```diff
@@ -119,7 +119,10 @@ void TIMER_Power(){
 
     elevator = 2 == mode ? abs(CHAN_ReadInput(INP_THROTTLE)) : abs(CHAN_ReadInput(INP_ELEVATOR));
     new_throttle = 2 == mode ?  abs(CHAN_ReadInput(INP_ELEVATOR)) : abs(CHAN_ReadInput(INP_THROTTLE));
-    new_throttle = abs(new_throttle - throttle);
+    if (new_throttle > throttle)
```
howard0su: this is warning that both are unsigned and result is unsigned as well. so we have to explict check to avoid use abs, which doesn't work.

PhracturedBlue: yes, this one looks good

src/pages/common/_model_loadsave.c
PhracturedBlue: idx can be an int, but the cast should be to an equivalent sized pointer I think. intptr_t would work for the cast.(I didn't check, I assume we have both intptr_t and uintptr_t)

src/telemetry/telem_crsf.c
howard0su: i would like to leave strcpy here for now. We should cleanup telemetry code later.

PhracturedBlue: Lint Error:
```
92:  Almost always, snprintf is better than strcpy  
```

src/protocol/inav_nrf24l01.c
goebish: I suppose that was a typo, 4 is fine for EMULATOR

howard0su: no one knows why. but goebish believe this is fine.

src/telemetry/telem_crsf.c
PhracturedBlue: Lint Error:
```
92:  Almost always, snprintf is better than strcpy  
```

PR #722, opened 2019-02-20
STM32 reorg: Remove hard-coded I2C values for Devof12e

PhracturedBlue:
add wrapper functions for i2c

### DIFF
--- a/src/target/drivers/mcu/stm32/f1/freq.h
+++ b/src/target/drivers/mcu/stm32/f1/freq.h
@@ -1,0 +1,10 @@
+#ifndef _DTX_STM32F1_FREQ_H_
+#define _DTX_STM32F1_FREQ_H_
+
+#define AHB_FREQ_MHz  72
+#define APB1_FREQ_MHz 36
+#define APB2_FREQ_MHz 72
+#define ADC_FREQ_MHz   9
+#define FREQ_MHz      AHB_FREQ_MHz
+
+#endif
--- a/src/target/drivers/mcu/stm32/i2c.h
+++ b/src/target/drivers/mcu/stm32/i2c.h
@@ -35,7 +35,6 @@ static inline void _i2c_init(struct i2c_config i2c)
     GPIO_setup_output_af(i2c.scl_sca, OTYPE_OPENDRAIN, i2c.i2c);
 
     i2c_peripheral_enable(i2c.i2c);
-    
 }
 
 #endif  // _DTX_STM32_I2C_H_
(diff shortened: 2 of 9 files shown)


src/target/drivers/mcu/stm32/i2c.h
```diff
@@ -0,0 +1,41 @@
...
+        ccr = (1000000 * APB1_FREQ_MHz) / (2 * i2c.i2c_freq);
+    }
+    i2c_set_ccr(i2c.i2c, 0xfff & ccr);
+    i2c_set_dutycycle(i2c.i2c, i2c.dutycycle);
+    i2c_set_trise(i2c.i2c, max_trise + 1);
+
+    GPIO_setup_output_af(i2c.scl_sca, OTYPE_OPENDRAIN, i2c.i2c);
+
+    i2c_peripheral_enable(i2c.i2c);
+
```
PhracturedBlue: Lint Error:
```
38:  Line ends in whitespace.  Consider deleting these extra spaces.  
38:  Redundant blank line at the end of a code block should be deleted.  
```

src/target/drivers/mcu/stm32/f1/freq.h
```diff
@@ -0,0 +1,10 @@
+#ifndef _DTX_STM32F1_FREQ_H_
+#define _DTX_STM32F1_FREQ_H_
+
+#define AHB_FREQ_MHz  72
+#define APB1_FREQ_MHz 36
+#define APB2_FREQ_MHz 72
+#define ADC_FREQ_MHz   9
+#define FREQ_MHz      AHB_FREQ_MHz
```
howard0su: should we define FREQ_MHZ and derive the other value from this?

PhracturedBlue: You can't do that.  the values can be set arbitrarily during clock config.  What we do need to do is to validate that the settings in clock config match our defines.  These values all exist as variables, but they aren't constants in the code, and so can't be optimized.  Since we know the values, defining them as constants simplifies some things

howard0su: okay, but when FREQ_MHZ is fixed to AHB_FREQ? I assume FREQ_MHZ is CPU clock while AHB_FREQ is some bus clock. right?

PhracturedBlue: FREQ_MHz is the core CPU frequency.  This is defined as the AHB Frequency at least on the F1.

PR #728, opened 2019-02-20
Fix SPI bug from #716

PhracturedBlue:


src/target/drivers/mcu/stm32/spi.h
```diff
@@ -12,9 +12,9 @@ static void _spi_init(struct spi_config spi_cfg)
         GPIO_setup_output_af(spi_cfg.mosi, OTYPE_PUSHPULL, spi_cfg.spi);
     }
     if (HAS_PIN(spi_cfg.miso)) {
-        GPIO_setup_input(spi_cfg.mosi, ITYPE_FLOAT);
+        GPIO_setup_input(spi_cfg.miso, ITYPE_FLOAT);
     }
-    spi_reset(spi_cfg.spi);
+    // spi_reset(spi_cfg.spi);
```
howard0su: why comment out this?

PhracturedBlue: we've never called spi_reset in the past.  I think it isn't necessary. Leaving it commented until I'm sure

PR #741, opened 2019-02-21
Support write for the flash inside MCU

howard0su:
on AT9 and AT10, we are using MCU flash to simulate a SPI flash. Enable the write support for the disk.


src/battery.c
```diff
@@ -45,12 +45,12 @@ u8 BATTERY_Check()
         PAGE_Test();
         CONFIG_SaveModelIfNeeded();
         CONFIG_SaveTxIfNeeded();
-        SPI_FlashBlockWriteEnable(0); //Disable writing to all banks of SPIFlash
+        Flash_BlockWriteEnable(0);  // Disable writing to all banks of SPIFlash
         warned |= BATTERY_CRITICAL;
```
PhracturedBlue: How about STORAGE_BlockWriteEnable instead?  these same routines potentially apply to SD_Card too (whick I know is still 'flash' but we don't call it that)

src/main.c
```diff
@@ -133,8 +133,8 @@ int main() {
     if(PWR_CheckPowerSwitch()) PWR_Shutdown();
 #if SPI_BOOTLOADER
     Initialize_ButtonMatrix();
-    SPIFlash_Init(); //This must come before LCD_Init() for 7e
-    SPI_FlashBlockWriteEnable(1); //Enable writing to all banks of SPIFlash
+    Flash_Init();  // This must come before LCD_Init() for 7e
+    Flash_BlockWriteEnable(1);  // Enable writing to all banks of SPIFlash
```
PhracturedBlue: STORAGE_Init

PR #745, opened 2019-02-22
STM32 reorg: remove hardcoded gpio pin definitions

PhracturedBlue:
Replace all hardcoded GPIO pins with those defined in hardware.h

### DIFF
--- a/src/target/drivers/input/rotary/rotary.c
+++ b/src/target/drivers/input/rotary/rotary.c
@@ -1,9 +1,21 @@
-#include <libopencm3/stm32/rcc.h>
-#include <libopencm3/stm32/gpio.h>
-#include <libopencm3/cm3/nvic.h>
-#include <libopencm3/stm32/exti.h>
+/*
+ This project is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
 
+ Deviation is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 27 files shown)


src/target/drivers/mcu/stm32/f1/pwr.h
```diff
@@ -0,0 +1,17 @@
+#ifndef _DTX_STM32F1_PWR_H_
+#define _DTX_STM32F1_PWR_H_
+
+static inline void _pwr_init()
+{
+    SCB_VTOR = VECTOR_TABLE_LOCATION;
+    SCB_SCR  &= ~SCB_SCR_SLEEPONEXIT; //sleep immediate on WFI
```
PhracturedBlue: Lint Error:
```
7:  At least two spaces is best between code and comments  
7:  Should have a space between // and comment  
```

src/target/drivers/input/rotary/rotary.c
```diff
@@ -14,26 +26,34 @@ static u32 last_rotary_clock = 0;
 
 void ROTARY_Init()
 {
-    rcc_peripheral_enable_clock(&RCC_APB2ENR, RCC_APB2ENR_IOPCEN);
+    rcc_periph_clock_enable(get_rcc_from_pin(ROTARY_PIN0));
+    rcc_periph_clock_enable(get_rcc_from_pin(ROTARY_PIN1));
+    rcc_periph_clock_enable(get_rcc_from_pin(ROTARY_PRESS_PIN));
```
howard0su: this is redundant? can we creates a set of macro to make hardware.h looks like a table close to what we have in our wiki?

and a code to scan the table to initialize IO.
and a header file to expose the right macros to the rest of code? like pagelist.h and page.h we are doing?

PhracturedBlue: you're right.  I have an assert to ensure they're the same already.  no need t do it 3 times

PR #750, opened 2019-02-22
Abstract ADC definitions

PhracturedBlue:


src/target/drivers/mcu/stm32/adc.c
```diff
@@ -19,15 +19,60 @@
 #include "target/drivers/mcu/stm32/dma.h"
 #include "target/drivers/mcu/stm32/rcc.h"
 
-#include "target/tx/devo/common/devo.h"  // FIXME - Shouldn't rely on things from 'devo' here
+#define NUM_ADC_CHANNELS (INP_HAS_CALIBRATION + 2)  // Inputs + Temprature + Voltage
+#define WINDOW_SIZE 10
+#define SAMPLE_COUNT NUM_ADC_CHANNELS * WINDOW_SIZE * ADC_OVERSAMPLE_WINDOW_COUNT
 
 unsigned ADC_Read(unsigned channel);
 volatile u16 adc_array_raw[NUM_ADC_CHANNELS];
```
howard0su: we need a better place to declare this. it is current only in devo.h

PR #764, opened 2019-02-24
Switch to libopencm3 usb stack

howard0su:
Use libopencm3 as  the usb stack to implement a new MSC protocol.


src/target/drivers/storage/mcu_flash.c
```diff
@@ -86,5 +84,8 @@ void MCUFlash_EraseSector(u32 sectorAddress) {
     if (!writable)
         return;
 
+    // Erase two pages in one call
+    // we simulate a 4KB based device
     flash_erase_page(sectorAddress);
+    flash_erase_page(sectorAddress + FLASH_PAGE_SIZE);
 }
```
PhracturedBlue: I think this should go into EraseSector4k() or something.  It is non intuitive that it will erase 2 sectors.  What if we set devofs sector size to 2k in the future (which is really what it should be on the at9/10)?

src/target/drivers/usb/devo_usb.c
```diff
@@ -0,0 +1,76 @@
...
+    .bcdDevice = 0x0200,
+    .iManufacturer = 1,
+    .iProduct = 2,
+    .iSerialNumber = 3,
+    .bNumConfigurations = 1,
+};
+
+const char USB_Product_Name[] = "DeviationTx";
+const char USB_Serial_Number[] = "5.0";
+
```
PhracturedBlue: I don't like hardcoding this version here.  I'll never remember to change it.  can we use the version-string instead?

src/target/drivers/storage/mcu_flash.c
howard0su: will address in a seperate PR to abstract out Disk

src/target/drivers/usb/devo_usb.c
howard0su: fixed

PR #770, opened 2019-02-26
Support LTO build

howard0su:
For any non-modular build Tx, you can use the following command to build Tx with LTO enabled:
ENABLE_LTO=1 make -j4 XXXXXXX.

LTO is enabled for devo7e by default.


src/target/tx/devo/common/Makefile.inc
```diff
@@ -82,6 +85,16 @@ ALL = $(LIBOPENCM3) $(TARGET).dfu
...
+CFLAGS_pwm_isr = -fno-lto
+CFLAGS_err_isr = -fno-lto
+CFLAGS_clock_isr = -fno-lto
+CFLAGS_usb_isr = -fno-lto
+CFLAGS_ppmin_isr = -fno-lto
+CFLAGS_uart_isr = -fno-lto
+CFLAGS_soft_serial_isr = -fno-lto
+endif
+
 ifeq ($(OS),Windows_NT)
```
PhracturedBlue: How about something like:
```
$(foreach src,$(filter %_isr.c,$(SRCS)),\
            $(eval CFLAG_$(src) = -fno-lto))
```

PhracturedBlue:
what about libopencm3 lto?

src/target/tx/devo/common/Makefile.inc
howard0su: sure. I like this solution. but not sure how i should use this in makefile.

howard0su:
> what about libopencm3 lto?

I would wait for libopencm3 to update to LTO.

src/target/tx/devo/common/Makefile.inc
PhracturedBlue: I think it will just work as is?  Probably need to be tweaked to look like:
```
$(foreach src,$basename $(notdir $(filter %_isr.c,$(SRCS)))),\
            $(eval CFLAG_$(src) = -fno-lto))
```
Try just using this as is instead of your current CFLAG definitions

PR #774, opened 2019-02-28
Add inverted ppmout option

goebish:
Add an option to reverse signal polarity to the PPM protocol.
~~Note that PPM out is broken in current master (at least on T8SG V2+), it outputs only 1 pulse per frame, maybe a DMA issue ?~~ (fixed by #777 )
Tested with an older version.

### DIFF
--- a/src/protocol/ppmout.c
+++ b/src/protocol/ppmout.c
@@ -38,6 +38,7 @@ static const char * const ppm_opts[] = {
   _tr_noop("Delta PW"),   "100",   "700",   STEP_SIZE, NULL,
   _tr_noop("Notch PW"),   "100",   "500",   STEP_SIZE, NULL,
   _tr_noop("Frame Size"), "10000", "30000", STEPSIZE2, NULL,
+  _tr_noop("Polarity"), _tr_noop("Uninverted"), _tr_noop("Inverted"), NULL,
   NULL
 };
 enum {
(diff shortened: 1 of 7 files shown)


src/protocol/ppmout.c
```diff
@@ -38,13 +38,15 @@ static const char * const ppm_opts[] = {
   _tr_noop("Delta PW"),   "100",   "700",   STEP_SIZE, NULL,
   _tr_noop("Notch PW"),   "100",   "500",   STEP_SIZE, NULL,
   _tr_noop("Frame Size"), "10000", "30000", STEPSIZE2, NULL,
+  _tr_noop("Polarity"), _tr_noop("Uninverted"), _tr_noop("Inverted"), NULL,
```
howard0su: I suggest use Normal as the default setting's name.

src/protocol/ppmout.c
```diff
@@ -66,7 +68,7 @@ static void build_data_pkt()
 static u16 ppmout_cb()
 {
     build_data_pkt();
-    PPM_Enable(Model.proto_opts[NOTCH_PW], pulses, Model.num_channels+1);
+    PPM_Enable(Model.proto_opts[NOTCH_PW], pulses, Model.num_channels+1, Model.proto_opts[POLARITY]);
```
howard0su: we need handle proto_opts change.

goebish: That's not worth it as it's set every time the callback is called.

PR #780, opened 2019-03-01
Fix protocol init when exit "Scanner" page with scanner=OFF

vladislavy:


src/pages/common/_scanner_page.c
```diff
@@ -139,5 +139,6 @@ void PAGE_ScannerEvent()
 
 void PAGE_ScannerExit()
 {
-    _scan_enable(0);
+    if(sp->enable)
```
PhracturedBlue: Lint Error:
```
142:  Missing space before ( in if(  
```

PR #776, opened 2019-03-01
Fix rare nrf24l01 problem

goebish:
Fix rare nrf24l01 problem where CS goes HIGH before the last clock pulse has completed:
fail
When this happens, the last byte is not registered by the nrf24l01 and is replaced by the previous one.

PR #783, opened 2019-03-02
Add protocol for Eachine E016H, from goebish

vladislavy:
Channel mapping:
Channel 5: flip
Channel 6: headless
Channel 7: RTH
Channel 8: emergency stop


goebish:
This protocol requires this patch to work: https://github.com/DeviationTX/deviation/pull/776

goebish:
The problem was recently introduced (and fixed), you can merge this protocol into 5.0 without any other changes.

PR #784, opened 2019-03-02
Add Silverware's analog aux channel feature to Bayang protocol

brianquad:
This change to the Bayang protocol adds an option to steal two presently unused (by the Silverware firmware, which targets several Bayang quads) in the protocol to send two more "analog" (non-binary) channels of data to the Rx. With corresponding changes on the Rx firmware, those new aux channels can be used to do more interesting things than the other binary aux channels. Some users have really liked PID tuning with the extra channels.

More information on the Silverware additions here:
https://www.rcgroups.com/forums/showthread.php?3121896-Analog-Aux-Channels-for-Silverware

The main NFE Silverware fork now has this feature available:
https://github.com/NotFastEnuf/NFE_Silverware

I am making a corresponding pull request to DIY Multiprotocol TX with the same additions to the Bayang protocol.

### DIFF
--- a/src/protocol/bayang_nrf24l01.c
+++ b/src/protocol/bayang_nrf24l01.c
@@ -67,6 +67,8 @@ enum {
     CHANNEL10,                  // Return To Home
     CHANNEL11,                  // Take Off/Landing
     CHANNEL12,                  // Emg. stop
+    CHANNEL13,                  // Analog Aux1
+    CHANNEL14,                  // Analog Aux2
 };
 #define CHANNEL_INVERTED    CHANNEL5    // inverted flight on Floureon H101
 #define CHANNEL_FLIP        CHANNEL6
@@ -76,6 +78,8 @@ enum {
 #define CHANNEL_RTH         CHANNEL10
 #define CHANNEL_TO          CHANNEL11
 #define CHANNEL_EMGSTOP     CHANNEL12
+#define CHANNEL_ANAAUX1     CHANNEL13
+#define CHANNEL_ANAAUX2     CHANNEL14
 enum {
     Bayang_INIT1 = 0,
     Bayang_BIND2,
@@ -85,6 +89,7 @@ enum {
 static const char *const bay_opts[] = {
     _tr_noop("Telemetry"), _tr_noop("Off"), _tr_noop("On"), NULL,
     _tr_noop("Format"), _tr_noop("regular"), "X16-AH", "IRDRONE", NULL,
+    _tr_noop("Analog Aux"), _tr_noop("Off"), _tr_noop("On"), NULL,
     NULL
 };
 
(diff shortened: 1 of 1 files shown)


src/protocol/bayang_nrf24l01.c
```diff
@@ -191,15 +204,20 @@ static void send_packet(u8 bind)
                     packet[0] = 0xa6;
                     break;
         }
-        packet[1] = 0xfa;       // normal mode is 0xf7, expert 0xfa
+        if (analogaux) {
+            packet[1] = scale_channel(CHANNEL_ANAAUX1, 0, 0xff);
+        }
+        else {
```
PhracturedBlue: Lint Error:
```
210:  If an else has a brace on one side, it should have it on both  
```

src/protocol/bayang_nrf24l01.c
```diff
@@ -217,7 +235,12 @@ static void send_packet(u8 bind)
     switch (Model.proto_opts[PROTOOPTS_FORMAT]) {
             case FORMAT_REGULAR:
                 packet[12] = txid[2];
-                packet[13] = 0x0a;
+                if (analogaux) {
+                    packet[13] = scale_channel(CHANNEL_ANAAUX2, 0, 0xff);
+                }
+                else {
```
PhracturedBlue: Lint Error:
```
241:  If an else has a brace on one side, it should have it on both  
```

src/protocol/bayang_nrf24l01.c
```diff
@@ -508,9 +534,9 @@ uintptr_t Bayang_Cmds(enum ProtoCmds cmd)
         initialize();
         return 0;
     case PROTOCMD_NUMCHAN:
-        return 12;
+        return (analogaux ? 14 : 12);
```
goebish: This should always be set to 14

brianquad: Thanks. I should have looked closer to figure out if that should change or not... PROTOCMD_DEFAULT_NUMCHAN is okay to change with the feature, though?

goebish: I think it will always return 12 because when creating is a new model analogaux == 0
Maybe better set it to 14 too ... and I4ll have to update the documentation

brianquad: Okay. Will returning 14 on both of these cause any problems for when analog aux is not enabled (Referring to two channels that don't exist)?

goebish: if analogaux is disabled then they will never be read, and they would just return 0 anyway.

goebish:
Do not merge for now, there's some issues with the Bayang protocol, better not add stuffs until it's fixed.
https://www.deviationtx.com/forum/3-feedback-questions/8319-can-not-bind-with-the-bayang-protocol#73159

goebish:
Problem fixed, ready to merge after one more dev review.

PR #786, opened 2019-03-03
[WIP] Refector Makefile

howard0su:


src/Makefile
```diff
@@ -4,10 +4,14 @@ unexport LANGUAGE
...
+mkfile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
+SDIR = $(dir $(mkfile_path))
+export
+
 ###########################################
 ## Provide some convenient make targets  ##
 ###########################################
-ifndef  TARGET
 TARGET   ?= devo8
 default: $(TARGET)
```
PhracturedBlue: We should get rid of the default.  running make without a target should just fail.

src/Makefile
```diff
@@ -32,380 +36,42 @@ everything: txs emus fss
...
-##################################################################################
-$(VERBOSE).SILENT:
-
-.PHONY: clean distclean language lint
-
-clean:
-	rm -f $(TARGET).$(EXEEXT) $(TARGET).exe $(TARGET).bin $(TARGET).dfu $(TARGET).list \
-		$(TARGET).map $(ODIR)/*.o $(ODIR)/*.o_ $(ODIR)/*.P  $(ODIR)/*.bin \
-		filesystem/$(FILESYSYTEM) 2> /dev/null || true
+$(foreach t,$(ALLTARGETS),$(eval $(call make-target,$t,$(notdir $(patsubst %/,%,$(dir $(wildcard target/tx/*/$(TARGET))))))))
```
PhracturedBlue: This doesn't seem right...how does $TARGET get set here?

src/Makefile.inc
```diff
@@ -0,0 +1,340 @@
...
+SRC_C    +=  $(wildcard $(SDIR)/firmware/*.c) \
+             $(wildcard $(SDIR)/misc/*.c) \
+             $(wildcard $(SDIR)/screen/*.c) \
+             $(wildcard $(SDIR)/screen/$(SCREENSIZE)/*.c) \
+             $(wildcard $(SDIR)/gui/*.c) \
+             $(wildcard $(SDIR)/pages/$(SCREENSIZE)/*.c) \
+             $(wildcard $(SDIR)/config/*.c)
+
+ifdef MODULAR
+   SRC_C    += $(SDIR)/protocol/protocol.c
```
PhracturedBlue: I never liked this.  Modular builds are a target thing, and don't belong in the main makefile.  but it is weird because the protocol code itself is here.  I'm not sure what to do about it.

src/Makefile.inc
```diff
@@ -0,0 +1,340 @@
...
+# recompile if the Makefile changes#
+####################################
+#$(OBJS) $(PROTO_OBJS) ($PROTO_EXTRA_OBJS): Makefile target/tx/$(FAMILY)/$(TARGET)/Makefile.inc $(EXTRA_MAKEFILES)
+
+##################################################################################
+# The following enables quiet output unless you use VERBOSE=1                    #
+# Note that this must be after the 1st rule so that it doesn't execute by default#
+##################################################################################
+$(VERBOSE).SILENT:
+
```
PhracturedBlue: This is one of the thinks I hate about Makefiles.  why does the order of things like this matter?  it is completely non-intuitive

src/Makefile.inc
```diff
@@ -0,0 +1,340 @@
...
+		filesystem/$(FILESYSYTEM) 2> /dev/null || true
+
+language:
+	../utils/extract_strings.pl -po > $(SDIR)/fs/language/deviation.po
+
+
+zip: $(TARGET).zip
+
+%.zip: $(ALL) $(TARGET).dfu $(PROTO_MODULES)
+	#This is not an emulator build (emulator is hanled in target/common/emu/Makefile.inc)
```
PhracturedBlue: knowledge of anything dfu related is target based and shouldn't be in the main file....but tis the general zip logic common to all tx?  it is different for emus....it is at least somewhat different for radiolnk and opentx. Not sure if we can break this up or move it without making it either duplicated or hard to read.

src/Makefile.inc
```diff
@@ -0,0 +1,340 @@
...
+	@mkdir -p $@
+
+$(MODELDIR):
+	@mkdir -p $@
+
+######################
+#The main executable #
+######################
+$(TARGET).$(EXEEXT): $(LINKFILE) $(OBJS) $(LIBOPENCM3)
+	@echo " + Building '$@'"
```
PhracturedBlue: libopencm3 stuff is target related.  it really doesn't belong in the main makefile.  except that it currently lives in the src/ directory.  maybe it should go into a libs/ directory or under target somewhere...I dunno

src/Makefile.inc
```diff
@@ -0,0 +1,340 @@
+
+# We want to turn on INCLUDE_FS by default
+INCLUDE_FS ?= 1
+
+# Use VERBOSE=1 to enable verbose make
+PROGMODE ?= STATUS_SCREEN
+
```
PhracturedBlue: What is this?  I can't find any reference to it anywhere else.  No idea why we have it.

src/Makefile.inc
howard0su: sure. we can. we need first define the MACROs and document somewhere. then the code can be much clean.

src/Makefile.inc
howard0su: i don't know.

src/Makefile.inc
```diff
@@ -213,31 +215,30 @@ $(MODELDIR):
...
 $(SDIR)$(TARGET).$(EXEEXT): $(LINKFILE) $(OBJS) $(LIBOPENCM3)
-	@echo " + Building '$@' '$(SRC_CXX)'"
+	@echo " + Building '$(notdir $@)'"
 ifdef LINKFILE #Create an empty 'obj/$(TARGET)/optimize.ld' just in case the linker script needs it
 	echo "" > $(ODIR)optimize.ld
 endif
 ifeq ("$(SRC_CXX)", "")
-	$(CC) -o $@ $(OBJS) $(LIBOPENCM3) $(LFLAGS) $(CFLAGS) $(EXTRA_CFLAGS)
+	$(CC) -o $@ $(OBJS) $(LIBOPENCM3) $(LFLAGS) $(LFLAGS2) $(CFLAGS) $(EXTRA_CFLAGS)
 else
```
PhracturedBlue: Can we name his better? I have no idea what the difference is between LFLAGS and LFLAGS2 (Yes I know you didn't create it, but now is a good time to fix it)

src/target/tx/devo/devo10/Makefile
```diff
@@ -8,4 +8,7 @@ LANGUAGE   := devo10
 
 OPTIMIZE_DFU := 1
 
+TARGET=devo10
+
```
PhracturedBlue: Why do I need this in every Makefile?  Isn't it inferrable by the directory I'm in?  The less things I need to define the better.
On the other hand, hiding away definitions makes things harder to read...I'm torn

src/Makefile
```diff
@@ -18,8 +18,8 @@ ALLTARGETS = $(filter-out common,$(notdir $(wildcard target/tx/*/*)))
 TXS ?= devo6 devo7e devo7e-256 devo8 devo10 devo12e devo12 devof7 devof7-XMS devof4 devof4-XMS devof12e devof12e-XMS at9 at10 t8sg t8sg_v2 t8sg_v2_plus ir8m
 
 #Filter non-existant emus
-EMUS = $(foreach dir,$(TXS:%=emu_%),$(if $(wildcard target/$(dir)),$(dir),))
-ALLEMUS = $(foreach dir,$(ALLTXS:%=emu_%),$(if $(wildcard target/$(dir)),$(dir),))
+EMUS = $(addprefix emu_, $(TXS))
+ALLEMUS = $(addprefix emu_, $(ALLTXS))
```
PhracturedBlue: Does this work if not every target has an emu?

src/target/tx/devo/emu_devof7/Makefile
```diff
@@ -2,5 +2,11 @@ SCREENSIZE  := text
 NUM_MODELS  ?= 10
 FILESYSTEMS := common text
 
+TARGET = emu_devof7
+FAMILY = devo
 
-include target/drivers/mcu/emu/Makefile.inc
+SRC += $(SDIR)/target/tx/devo/devof7/ia9211_map.c
+CFLAGS +=-I$(SDIR)/target/drivers/display/spi/
```
PhracturedBlue: This is weird to me.  the ia9211_map file is emu only but lives in the devof7 dir?

src/Makefile
howard0su: I need to filter out some targets. but since we put target under family, it is a challenge to do so.

src/target/tx/devo/devo10/Makefile
howard0su: I struggled as well.
I can pass TARGET from Makefile, maybe I should do that.

src/target/tx/devo/emu_devof7/Makefile
howard0su: no. it is a include file used by devof7 lcd driver as well.

PR #793, opened 2019-03-04
Make empty string as the power for modules without RF

howard0su:


howard0su:
This is reported by user on forum as well.

PR #794, opened 2019-03-04
Cleanup _tr usage

howard0su:
Make most _tr calls only happen inside pages with the following exceptions:
1. input.c
2. telemetry
which will be addressed separately.

### DIFF
--- a/src/inputs.c
+++ b/src/inputs.c
@@ -89,28 +89,24 @@
 #define INPNAME_SWH0(x,y)      x = _tr_noop("SW H"); y = 0
 #define INPNAME_SWH1(x,y)      x = _tr_noop("SW H"); y = 1
 
-#define SWITCH_NAME_GEAR0 _tr("GEAR")
-#define SWITCH_NAME_GEAR1 _tr("GEAR")
-#define SWITCH_NAME_GEAR0 _tr("GEAR")
-
-#define BUTNAME_TRIM_LV_NEG _tr("TRIMLV-")
-#define BUTNAME_TRIM_LV_POS _tr("TRIMLV+")
-#define BUTNAME_TRIM_RV_NEG _tr("TRIMRV-")
-#define BUTNAME_TRIM_RV_POS _tr("TRIMRV+")
-#define BUTNAME_TRIM_LH_NEG _tr("TRIMLH-")
-#define BUTNAME_TRIM_LH_POS _tr("TRIMLH+")
-#define BUTNAME_TRIM_RH_NEG _tr("TRIMRH-")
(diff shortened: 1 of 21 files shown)


howard0su:
prepare to enable dynamic loc strings

howard0su:
Enough for this PR. I will start another two to clean up input.c and telemetry.c. I bet I can save another 100 bytes.

src/mixer_standard.c
```diff
@@ -164,9 +164,15 @@ unsigned STDMIXER_ValidateTraditionModel()
...
+{
+    _tr_noop("Advanced"),
+    _tr_noop("Standard"),
+};
+
 const char *STDMIXER_ModeName(int mode)
 {
-    return mode == MIXER_ADVANCED ? _tr("Advanced") : _tr("Standard");
+    return ModeNames[mode - 0x01];
 }
```
PhracturedBlue: I like the ternary better.  This has the potential for a buffer overrun if an invalid mode is passed

src/inputs.c
```diff
@@ -89,28 +89,24 @@
 #define INPNAME_SWH0(x,y)      x = _tr_noop("SW H"); y = 0
 #define INPNAME_SWH1(x,y)      x = _tr_noop("SW H"); y = 1
 
-#define SWITCH_NAME_GEAR0 _tr("GEAR")
-#define SWITCH_NAME_GEAR1 _tr("GEAR")
-#define SWITCH_NAME_GEAR0 _tr("GEAR")
-
```
PhracturedBlue: Why was GEAR removed?

src/rtc.c
```diff
@@ -308,9 +308,15 @@ void RTC_GetTimeGTM(struct gtm *t)
...
+    _tr_noop("Clock"),
+    _tr_noop("Date"),
+};
 const char *RTC_Name(char *str, int i)
 {
-    sprintf(str, "%s", i == 0 ? _tr("Clock") : _tr("Date"));
-    return str;
+    (void)str;
+    return RTCNames[i];
 }
```
PhracturedBlue: again, the ternary is safer.  if you don't want to use it, you need to validate that 'i' isn't out of bounds

src/inputs.c
howard0su: not used anywhere

src/mixer_standard.c
howard0su: okay

src/rtc.c
howard0su: okay

src/inputs.c
howard0su: I tracked the history, it is added by you but it is never used.

PR #795, opened 2019-03-04
Allow choosing primary UART interface for logging

PhracturedBlue:


src/target/drivers/serial/usb_cdc/usb_uart.c
```diff
@@ -413,3 +413,39 @@ void usb_vcp_init(void) {
...
+        if (CLOCK_getms() - lastms < 250) {
+            continue;
+        }
+        lastms = CLOCK_getms();
+        printf("hello\n");
+        gpio_toggle(GPIOC, GPIO4);
+    }
+}
+
+void UART_Initialize()
```
howard0su: shall we consolidate usb initialize?

PhracturedBlue: probably.  I'm not sure how to best do that in this case, but if you have some idea of how it should look, let me know

PR #796, opened 2019-03-04
Fix MMC access in Font code.  Add MMC drivers

PhracturedBlue:


src/screen/font.c
```diff
@@ -97,8 +97,10 @@ u8 open_font(const char* fontname)
     close_font();
 
     sprintf(filename, "media/%s.fon", fontname);
-    finit(&FontFAT, "media");
-    font.fh = fopen2(&FontFAT, filename, "rb");
+#ifdef FIL
```
howard0su: if FIL is not defined, how the line 19 get compiled?

src/target.h
```diff
@@ -217,8 +221,11 @@ int  PARFlash_ReadBytesStopCR(u32 readAddress, u32 length, u8 * buffer);
     #define STORAGE_ReadBytesStopCR MCUFlash_ReadBytesStopCR
     #define STORAGE_WriteBytes MCUFlash_WriteBytes
     #define STORAGE_EraseSector MCUFlash_EraseSector
+#elif FLASHTYPE == FLASHTYPE_NONE
+    #define STORAGE_WriteEnable(enable) if (0) {}
```
howard0su: use do {} while (0) instead.

src/target.h
```diff
@@ -217,8 +221,11 @@ int  PARFlash_ReadBytesStopCR(u32 readAddress, u32 length, u8 * buffer);
     #define STORAGE_ReadBytesStopCR MCUFlash_ReadBytesStopCR
     #define STORAGE_WriteBytes MCUFlash_WriteBytes
     #define STORAGE_EraseSector MCUFlash_EraseSector
+#elif FLASHTYPE == FLASHTYPE_NONE
```
howard0su: FLASH_TYPE_MMC?

src/screen/font.c
PhracturedBlue: FIL is always either a define or a typedef.  I really need a better way to abstract this.  the petitfat and devofs use a single handle for filesystem and filehandle.  FatFS uses seperate

howard0su: is finit is not defined in FatFS? why not define a dummy to avoid this ifdef?

src/target.h
```diff
@@ -217,8 +220,11 @@ int  PARFlash_ReadBytesStopCR(u32 readAddress, u32 length, u8 * buffer);
     #define STORAGE_ReadBytesStopCR MCUFlash_ReadBytesStopCR
     #define STORAGE_WriteBytes MCUFlash_WriteBytes
     #define STORAGE_EraseSector MCUFlash_EraseSector
+#elif FLASHTYPE == FLASHTYPE_MMC
+    #define STORAGE_WriteEnable(enable) do {} while(0);
```
PhracturedBlue: Lint Error:
```
224:  Missing space before ( in while(  
```

howard0su: please don't add ';' in the end

src/target/drivers/filesystems/syscalls.c
```diff
@@ -33,49 +33,47 @@ extern void printstr(char *, int);
 #define dbgMsec()          ((DEBUG_SYSCALLS) ? CLOCK_getms() : 0)
 
 static DIR   dir;
+
 #ifdef MEDIA_DRIVE
-static FATFS fat[2];
+static DRIVE drive[2];
```
howard0su: I would prefer to not have another magic number here. maybe we can move the definition from petit_io.c to here. but I am fine you fist get this in.

src/target/drivers/filesystems/syscalls.c
```diff
@@ -33,49 +33,47 @@ extern void printstr(char *, int);
...
 
 
-int FS_Mount(void *_f, const char *drive)
+int FS_Init()
 {
-    (void)drive;
-    if(! _f) {
 #ifdef MEDIA_DRIVE
-        int res = (FS_Mount(&fat[0], "")  && FS_Mount(&fat[1], "media"));
+    int res = (FS_Mount(&drive[0].fat, "")  && FS_Mount(&drive[1].fat, "media"));
```
howard0su: I would rather to define "media" as a const, it is referenced so many times in this file.

src/target/drivers/filesystems/syscalls.c
```diff
@@ -97,10 +95,10 @@ void FS_Unmount()
 
 int FS_OpenDir(const char *path)
 {
-    FATFS *ptr = &fat[0];
+    FATFS *ptr = &drive[0].fat;
```
howard0su: maybe this can use a MACRO to make the logic explicit like if (IN_MEDIA_DRIVER)? or
define a function:
 drive_num = get_drive_index_from_filename();
ptr = drive[drive_num].fat;

I don't really like this ifdef. it is not that special.

src/target/drivers/filesystems/syscalls.c
PhracturedBlue: I want to get rid of the entire MEDIA special case, but it requires passing a file-pointer to the filesystem layer.  I'm not going to make that change now.

src/target/drivers/filesystems/syscalls.c
PhracturedBlue: See above.  no reason to clean this further as the purpose is to get rid of it altogether

src/target/drivers/mcu/emu/stubs.c
```diff
@@ -96,11 +96,15 @@ void fempty(FILE *fh)
     fseek(fh, 0, SEEK_SET);
 }
 
+int FS_Init() {
+    printf("Changing directory to: '%s'\n", FILESYSTEM_DIR);
+    return ! chdir(FILESYSTEM_DIR);
+}
```
PhracturedBlue: Lint Error:
```
102:  Extra space for operator !   
```

PR #801, opened 2019-03-05
Enable LCD to init SPI when no SPI-Flash is present

PhracturedBlue:


src/target/drivers/display/spi/128x64x1_nt7538.c
```diff
@@ -87,6 +87,9 @@ void LCD_Init()
 {
     //Initialization is mostly done in SPI Flash
     //Setup CS as B.0 Data/Control = C.5
+#ifndef FLASH_SPI_CFG
+    _spi_init(LCD_SPI_CFG);
```
howard0su: what if we just let SPI initialize twice? we should not really assume the order to initialize.

PhracturedBlue: it isn't safe to initialize twice without a reset.  it would be better to have initialization happen in a central place.

PR #802, opened 2019-03-05
Move SPI intialization into F1 specific directory in preparation for …

PhracturedBlue:
…F2 merge

### DIFF
--- a/src/target/drivers/mcu/stm32/f1/spi.h
+++ b/src/target/drivers/mcu/stm32/f1/spi.h
@@ -3,7 +3,7 @@
 
 #include "rcc.h"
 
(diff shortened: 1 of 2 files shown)


src/target/drivers/mcu/stm32/f1/spi.h
```diff
@@ -0,0 +1,33 @@
+#ifndef _DTX_STM32F1_SPI_H_
+#define _DTX_STM32F1_SPI_H_
+
+#include "rcc.h"
+
+static void _spi_init(struct spi_config spi_cfg)
```
howard0su: inline

PR #807, opened 2019-03-06
Enable AFIO and ensure the CYRF6936 reset pin is configured

PhracturedBlue:
This should fix using Deviation with the new bootloader


src/target/tx/devo/common/spi_proto.c
```diff
@@ -92,8 +92,12 @@ void SPI_ProtoInit()
 {
 // If we use SPI Switch board then SPI2 is shared between RF chips
 // and flash, so it is initialized in SPIFlash.
-    if (HAS_PIN(PROTO_RST_PIN))
+    if (HAS_PIN(PROTO_RST_PIN)) {
```
howard0su: why not enable clock here?

PR #811, opened 2019-03-07
Fix devof12e toggles

vladislavy:


vladislavy:
Probably will be better to use GUI_Redraw(&gui->elem[i].box)

PR #812, opened 2019-03-08
New bar style for 128x64

howard0su:


PhracturedBlue:
Can you verify this on the horizontal trims?  It deosnt' seem to work properly:
image

I expect the trims to be unaffected by this patch, right?  the vertical trims seem to owkr as I'd expect

howard0su:
let me fix this before i reopen the PR

PR #813, opened 2019-03-09
Improve scanner code and move RF related stuff to protocol

TheRealMoeder:
Improvements:

1. move RF control to protocol 
2. improve scan time 
3. add optional 30dB attenuator 
4. make averaging adjustable between 1-8192
5. use exponential smoothing on averaging
6. add peak hold feature in averaging mode (currently only on monochrome displays)
7. support modular build


src/pages/128x64x1/scanner_page.c
```diff
@@ -55,12 +55,17 @@ static void _draw_page(u8 enable)
...
 {
     const unsigned offset = HEADER_HEIGHT + LINE_HEIGHT;
-    // draw a line
-    int col = (LCD_WIDTH - (MAX_RADIOCHANNEL - MIN_RADIOCHANNEL)) / 2 + sp->channel;
-    int height = sp->channelnoise[sp->channel] * (LCD_HEIGHT - offset) / 0x1F;
+    int col, height;
+    int num_chan = sp->chan_max - sp->chan_min;
+ 
+    // draw rssi values
+    for(int i = 0; i < num_chan; i++) {
```
PhracturedBlue: Lint Error:
```
60:  Weird number of spaces at line-start.  Are you using a 2-space indent?  
60:  Line ends in whitespace.  Consider deleting these extra spaces.  
62:  Missing space before ( in for(  
```

src/pages/128x64x1/scanner_page.c
```diff
@@ -55,12 +55,17 @@ static void _draw_page(u8 enable)
...
+        col = (LCD_WIDTH - (num_chan)) / 2 + i;
+        height = sp->rssi[i] * (LCD_HEIGHT - offset) / 0x1F;
 
-    LCD_DrawFastVLine(col, offset, LCD_HEIGHT - offset - height, 0);
-    LCD_DrawFastVLine(col, LCD_HEIGHT - height, height, Display.xygraph.grid_color);
-}
+        LCD_DrawFastVLine(col, offset, LCD_HEIGHT - offset - height, 0);
+        LCD_DrawFastVLine(col, LCD_HEIGHT - height, height, Display.xygraph.grid_color);
+    }
+ }
```
PhracturedBlue: Lint Error:
```
69:  Weird number of spaces at line-start.  Are you using a 2-space indent?  
```

src/protocol/scanner_cyrf6936.c
```diff
@@ -0,0 +1,156 @@
...
+};
+
+static void cyrf_init()
+{
+    /* Initialise CYRF chip */
+    CYRF_WriteRegister(CYRF_1D_MODE_OVERRIDE, 0x38);  //FRC SEN (forces the synthesizer to start) + FRC AWAKE (force the oscillator to keep running at all times)
+    CYRF_WriteRegister(CYRF_03_TX_CFG, 0x08 | 7);     //Data Code Length = 32 chip codes + Data Mode = 8DR Mode + max-power(+4 dBm)
+    CYRF_WriteRegister(CYRF_06_RX_CFG, 0x4A);         //LNA + FAST TURN EN + RXOW EN, enable low noise amplifier, fast turning, overwrite enable
+    CYRF_WriteRegister(CYRF_0B_PWR_CTRL, 0x00);       //Reset power control
+    CYRF_WriteRegister(CYRF_10_FRAMING_CFG, 0xA4);    //SOP EN + SOP LEN = 32 chips + LEN EN + SOP TH = 04h
```
PhracturedBlue: Lint Error:
```
40:  Should have a space between // and comment  
41:  Should have a space between // and comment  
42:  Should have a space between // and comment  
43:  Should have a space between // and comment  
44:  Should have a space between // and comment  
```

src/protocol/scanner_cyrf6936.c
```diff
@@ -0,0 +1,156 @@
...
+    CYRF_WriteRegister(CYRF_11_DATA32_THOLD, 0x05);   //TH32 = 0x05
+    CYRF_WriteRegister(CYRF_12_DATA64_THOLD, 0x0E);   //TH64 = 0Eh, set pn correlation threshold
+    CYRF_WriteRegister(CYRF_1B_TX_OFFSET_LSB, 0x55);  //STRIM LSB = 0x55, typical configuration
+    CYRF_WriteRegister(CYRF_1C_TX_OFFSET_MSB, 0x05);  //STRIM MSB = 0x05, typical configuration
+    CYRF_WriteRegister(CYRF_32_AUTO_CAL_TIME, 0x3C);  //AUTO_CAL_TIME = 3Ch, typical configuration
```
PhracturedBlue: Lint Error:
```
45:  Should have a space between // and comment  
46:  Should have a space between // and comment  
47:  Should have a space between // and comment  
48:  Should have a space between // and comment  
49:  Should have a space between // and comment  
```

src/protocol/scanner_cyrf6936.c
```diff
@@ -0,0 +1,156 @@
...
+    CYRF_WriteRegister(CYRF_35_AUTOCAL_OFFSET, 0x14); //AUTO_CAL_OFFSET = 14h, typical configuration
+    CYRF_WriteRegister(CYRF_39_ANALOG_CTRL, 0x01);    //ALL SLOW
+    CYRF_WriteRegister(CYRF_1E_RX_OVERRIDE, 0x10);    //FRC RXDR (Force Receive Data Rate)
+    CYRF_WriteRegister(CYRF_1F_TX_OVERRIDE, 0x00);    //Reset TX overrides
+    CYRF_WriteRegister(CYRF_01_TX_LENGTH, 0x10);      //TX Length = 16 byte packet
```
PhracturedBlue: Lint Error:
```
50:  At least two spaces is best between code and comments  
50:  Should have a space between // and comment  
51:  Should have a space between // and comment  
52:  Should have a space between // and comment  
53:  Should have a space between // and comment  
54:  Should have a space between // and comment  
```

src/protocol/scanner_cyrf6936.c
```diff
@@ -0,0 +1,156 @@
...
+    CYRF_WriteRegister(CYRF_27_CLK_OVERRIDE, 0x02);   //RXF, force receive clock
+    CYRF_WriteRegister(CYRF_28_CLK_EN, 0x02);         //RXF, force receive clock enable
```
PhracturedBlue: Lint Error:
```
55:  Should have a space between // and comment  
56:  Should have a space between // and comment  
```

src/protocol/scanner_cyrf6936.c
```diff
@@ -0,0 +1,156 @@
...
+}
+
+static void _scan_next()
+{
+    CYRF_ConfigRFChannel(sp->channel + sp->chan_min);
+    if(sp->attenuator) {
```
PhracturedBlue: Lint Error:
```
62:  Missing space before ( in if(  
```

src/protocol/scanner_cyrf6936.c
```diff
@@ -0,0 +1,156 @@
...
+        CYRF_WriteRegister(CYRF_06_RX_CFG, 0x4A);
+    }
+}
+
+static int _scan_rssi()
+{
+    if ( !(CYRF_ReadRegister(CYRF_05_RX_CTRL) & 0x80)) {
+        CYRF_WriteRegister(CYRF_05_RX_CTRL, 0x80); //Prepare to receive
+        Delay(10);
+        CYRF_ReadRegister(CYRF_13_RSSI); //dummy read
```
PhracturedBlue: Lint Error:
```
71:  Mismatching spaces inside () in if  
72:  At least two spaces is best between code and comments  
72:  Should have a space between // and comment  
74:  At least two spaces is best between code and comments  
74:  Should have a space between // and comment  
```

src/protocol/scanner_cyrf6936.c
```diff
@@ -0,0 +1,156 @@
...
+}
+
+static u16 scan_cb()
+{
+    int rssi_update;
+    switch (sp->scanState) {
+        case SCAN_CHANNEL_CHANGE:
+            averages = 0;
+            sp->channel++;
+            if(sp->channel == (sp->chan_max - sp->chan_min + 1))
```
PhracturedBlue: Lint Error:
```
91:  Missing space before ( in if(  
```

src/protocol/scanner_cyrf6936.c
```diff
@@ -0,0 +1,156 @@
...
+            sp->rssi[sp->channel] = 0;
+            _scan_next();
+            sp->scanState = SCAN_GET_RSSI;
+            return CHANNEL_LOCK_TIME;
+        case SCAN_GET_RSSI:
+            rssi_update = _scan_rssi();
+            if(sp->scan_mode) {
+                sp->rssi[sp->channel] = (sp->rssi[sp->channel] + rssi_update) / 2;
+            } else {
+                if(rssi_update > sp->rssi[sp->channel])
```
PhracturedBlue: Lint Error:
```
99:  Missing space before ( in if(  
102:  Missing space before ( in if(  
```

src/protocol/scanner_cyrf6936.c
```diff
@@ -0,0 +1,156 @@
...
+    sp->chan_max = MAX_RADIOCHANNEL;
+    CYRF_Reset();
+    cyrf_init();
+    CYRF_SetTxRxMode(RX_EN);  // Receive mode
+    scan_start();
+}
+
+uintptr_t SCANNER_CYRF_Cmds(enum ProtoCmds cmd)
+{
+    switch(cmd) {
```
PhracturedBlue: Lint Error:
```
132:  Missing space before ( in switch(  
```

src/pages/common/_scanner_page.c
```diff
@@ -99,18 +42,22 @@ static void press_enable_cb(guiObject_t *obj, const void *data)
...
 {
     (void)data;
-    sp->attenuator ^= 1;
-    GUI_Redraw(obj);
+    GUI_TextSelectEnablePress((guiTextSelect_t *)obj, sp->averaging);
+    sp->averaging = GUI_TextSelectHelper(sp->averaging, 0, 255, dir, 1, 10, NULL);
+    if (sp->averaging == 0)
+        strcpy(tempstring, _tr("Peak"));
+    else
+        snprintf(tempstring, 8, "Avg %d", sp->averaging);
```
PhracturedBlue: Lint Error:
```
59:  If you can, use sizeof(tempstring) instead of 8 as the 2nd arg to snprintf.  
```

src/protocol/scanner_cyrf6936.c
```diff
@@ -0,0 +1,162 @@
...
+
+#define MIN_RADIOCHANNEL    0x00
+#define MAX_RADIOCHANNEL    0x62
+#define CHANNEL_LOCK_TIME   300  // slow channel requires 270 usec for synthesizer to settle
+#define INTERNAL_AVERAGE         3
+#define AVERAGE_INTVL       50
+
+static struct scanner_page * const sp = &pagemem.u.scanner_page;
+static int averages;
+static long rssi_sum;
```
PhracturedBlue: Lint Error:
```
30:  Use int16/int64/etc, rather than the C type long  
```

src/pages/common/_scanner_page.c
```diff
@@ -99,18 +42,27 @@ static void press_enable_cb(guiObject_t *obj, const void *data)
...
```
PhracturedBlue: Lint Error:
```
64:  If you can, use sizeof(tempstring) instead of 8 as the 2nd arg to snprintf.  
```

PhracturedBlue: Lint Error:
```
64:  If you can, use sizeof(tempstring) instead of 8 as the 2nd arg to snprintf.  
```

src/pages/320x240x16/scanner_page.c
```diff
@@ -35,41 +35,28 @@ static const char *enablestr_cb(guiObject_t *obj, const void *data)
...
 
 void _draw_channels()
 {
-    GUI_Redraw(&gui->bar[sp->channel]);
+    if (!sp->bars_valid) {
+        const unsigned int height = LCD_HEIGHT - 78;
+        int width = LCD_WIDTH / (sp->chan_max - sp->chan_min + 1);
+        int xoffset = (LCD_WIDTH - width * ((sp->chan_max - sp->chan_min) + 1))/2;
+        for(int i = 0; i < (sp->chan_max - sp->chan_min); i++) {
+            GUI_CreateBarGraph(&gui->bar[i], xoffset + i * width, 70, width, height, 2, 31, BAR_VERTICAL, show_bar_cb, (void *)((long)i));
```
PhracturedBlue: Lint Error:
```
53:  Missing space before ( in for(  
54:  Use int16/int64/etc, rather than the C type long  
```

src/pages/common/_scanner_page.c
```diff
@@ -99,18 +42,27 @@ static void press_enable_cb(guiObject_t *obj, const void *data)
...
 {
+    (void)obj;
     (void)data;
-    sp->attenuator ^= 1;
-    GUI_Redraw(obj);
+    sp->averaging = GUI_TextSelectHelper(sp->averaging, 0, 255, dir, 1, 10, NULL);
+    if (sp->averaging == 0)
+        strcpy(tempstring, _tr("Peak"));
+    else
+        snprintf(tempstring, 8, "Avg %d", sp->averaging);
```
PhracturedBlue: Lint Error:
```
64:  If you can, use sizeof(tempstring) instead of 8 as the 2nd arg to snprintf.  
```

PhracturedBlue: Lint Error:
```
64:  If you can, use sizeof(tempstring) instead of 8 as the 2nd arg to snprintf.  
```

src/pages/common/_scanner_page.c
```diff
@@ -99,18 +42,27 @@ static void press_enable_cb(guiObject_t *obj, const void *data)
...
 }
 
-static void press_mode_cb(guiObject_t *obj, const void *data)
+static const char *attenuator_cb(guiObject_t *obj, int dir, void *data)
 {
+    (void)obj;
     (void)data;
-    sp->scan_mode ^= 1;
-    GUI_Redraw(obj);
+    sp->attenuator = GUI_TextSelectHelper(sp->attenuator, 0, 2, dir, 1, 1, NULL);
```
howard0su: can we use PROTO_OPTS to let the user choose the options? So that we don't need to create another set of UI elements. Also, this gives different options for different scanners.

src/pages/common/model_page.h
```diff
@@ -21,4 +21,10 @@ struct model_page {
     enum ModelType modeltype;
 };
 
+#if HAS_SCANNER
+#define PROTOCOL_SCANNERS 1
```
howard0su: PROTOCOL_SCANNERS always equals to HAS_SCANNER. why we need another constant?

src/pages/common/scanner_page.h
```diff
@@ -1,16 +1,17 @@
 #ifndef _SCANNER_PAGE_H_
 #define _SCANNER_PAGE_H_
 
-#define MIN_RADIOCHANNEL     0x00
-#define MAX_RADIOCHANNEL     0x4F
-
 struct scanner_page {
-    u8 channelnoise[MAX_RADIOCHANNEL - MIN_RADIOCHANNEL + 1];
+    u8 rssi[255];
```
howard0su: I don't like a magic number to define this. can we make it as a memory inside protocol? the other fields in this structure look like more options. let's split out them

src/protocol/protocol.h
```diff
@@ -78,4 +78,7 @@ PROTODEF(PROTOCOL_PXX, R9M,  NULL, PXXOUT_Cmds, "PXX")
 PROTODEF(PROTOCOL_TESTSER, TX_MODULE_LAST,  NULL, TESTSER_Cmds, "TESTSER")
 PROTODEF(PROTOCOL_TESTRF, TX_MODULE_LAST,  NULL, TESTRF_Cmds, "TESTRF")
 #endif
+#if HAS_SCANNER
```
howard0su: SUPPORT_SCANNER.

HAS_ is used to choose hardware. SUPPORT_ is used to define an optional software feature. But what's the reason we want to this feature be optional?

src/protocol/scanner_cyrf6936.c
```diff
@@ -0,0 +1,162 @@
...
+
+#ifdef PROTO_HAS_CYRF6936
+
+#define MIN_RADIOCHANNEL    0x00
+#define MAX_RADIOCHANNEL    0x62
+#define CHANNEL_LOCK_TIME   300  // slow channel requires 270 usec for synthesizer to settle
+#define INTERNAL_AVERAGE    3
+#define AVERAGE_INTVL       50
+
+static struct scanner_page * const sp = &pagemem.u.scanner_page;
```
howard0su: we should not reference pagemem from a protocol. please find a different way to do so.

src/pages/common/_scanner_page.c
TheRealMoeder: Proto opts would require leaving scanner page, but those I put in options need to be changed quickly and there effect immediately seen.

src/pages/common/model_page.h
TheRealMoeder: I planned on increasing PROTOCOL_SCANNERS to 2 when adding cc2500 scanner module.

src/pages/common/scanner_page.h
TheRealMoeder: You mean 255? It's maximum scan channels for cc2500, cyrf only had 90. Rssi has to be written to by protocol and read out by scanner page, I don't think we can move this memory to protocol (thinking devo7e)

src/protocol/protocol.h
TheRealMoeder: Agree on SUPPORT_. Making this optional until we know Rom and Ram impact on 7e

TheRealMoeder: Also, scanner doesn't work for text gui.

src/protocol/scanner_cyrf6936.c
TheRealMoeder: I leaft struct in pagemem since it already was there. Otherwise we could move it into transmitter struct (or other suggestions?). But this would likely increase memory usage.

src/pages/common/model_page.h
TheRealMoeder: you're right, I removed it.

src/pages/320x240x16/scanner_page.c
```diff
@@ -35,41 +35,28 @@ static const char *enablestr_cb(guiObject_t *obj, const void *data)
...
+            GUI_CreateBarGraph(&gui->bar[i], xoffset + i * width, 70, width, height, 2, 31, BAR_VERTICAL, show_bar_cb, (void *)(uintptr_t)i);
+        }
+        sp->bars_valid = 1;
+    }
+    for (int i = 0; i < (sp->chan_max - sp->chan_min); i++)
+        GUI_Redraw(&gui->bar[i]);
 }
 
-#endif //HAS_SCANNER
+#endif //SUPPORT_SCANNER
```
PhracturedBlue: Lint Error:
```
62:  At least two spaces is best between code and comments  
62:  Should have a space between // and comment  
```

src/pages/128x64x1/scanner_page.c
```diff
@@ -29,38 +29,29 @@ static const char *enablestr_cb(guiObject_t *obj, const void *data)
...
 
-    LCD_DrawFastVLine(col, offset, LCD_HEIGHT - offset - height, 0);
-    LCD_DrawFastVLine(col, LCD_HEIGHT - height, height, Display.xygraph.grid_color);
+        LCD_DrawFastVLine(col, offset, LCD_HEIGHT - offset - height, 0);
+        LCD_DrawFastVLine(col, LCD_HEIGHT - height, height, Display.xygraph.grid_color);
+    }
 }
 
-#endif //HAS_SCANNER
+#endif //SUPPORT_SCANNER
```
PhracturedBlue: Lint Error:
```
57:  At least two spaces is best between code and comments  
57:  Should have a space between // and comment  
```

vladislavy:
Previously "Average" mode was intended to show permanent signals only and filter out any burst transmissions. I dont's see sense to average a burst transmissions.

TheRealMoeder:
It all depends on the duty cycle of transmission compared to averaging time. Whether we do simple arithmetic average or exponentially smoothing measurements makes not too much difference except for how it responds to max power peaks or 0 measurements. By increasing averaging we can now see only signals with high duty cycle on a given frequency. Old code did averaging of 5 measurements with 5 samples in intervals of 300 us already, now we can increase that. 
Peak mode is actually currently not the same as before, it is just 0 averaging and showing instantaneous signal strength instead of a peak hold. Will have to change that. Please correct me if I am wrong.

TheRealMoeder:
Enabling scanner support for devo7e costs 670 bytes of ROM and 250 bytes of RAM.
I left it disabled for now, unless we agree it is wanted in nightlies.

vladislavy:
I'm not sure about probability of detection a burst transmissions, now we use about 300usec detection for single channel, next detection will be over 60msec. So, we have detection probability about 1/100.
I think we need at least 25msec on single channel for reliable burst transmission detection. Or, we need to use very fast channels scanning, but it's impossible for CYRF6936.

TheRealMoeder:
I thought you didn't want burst transmissions to be seen. What do you to think of changing the behavior to always make averaging (with variable iterations per channel) but also adding an optional peak hold? Or instead offering peak with the same iteration parameter as avg? So we can select from peak 1 to peak n and avg 1 to avg n?

vladislavy:
I think about packet transmission as a burst transmission, with duration about 1msec.
Also, I don't see any "Scanner.rssi" reset to 0 for Peak mode. How it will work?
And, I think will be better move "Scanner.rssi" reset to the end of screen refresh (// draw rssi values). 
memset(Scanner.rssi, 0, sizeof(Scanner.rssi));  // clear old rssi values

TheRealMoeder:
Thank you for your input. I actually removed scanner.rssi[c] = 0 for averaging in my local, as it causes some channels to flicker. I fear if we reset scanner.rssi for peak mode on every page event we will also get some ugly flicker, because scanning and drawing are not synchronized.

TheRealMoeder:
@vladislavy I rewrote peak and averaging modes including additional peak hold feature. I think it is a lot more usable now, also in revealing burst transmissions.

TheRealMoeder:
Channels are 2400 - 2498 MHz in 1 MHz spacing for the cyrf6936. Every pixel column on monochrome screen represents one channel ( one bargraph per channel for color screens).

PR #824, opened 2019-03-17
[WIP] Consolidate switch code 3/4

PhracturedBlue:
This will be a series of patches to remove channels.c. It is being broken down into pieces to make bisection easier in the case of regressions.

Part 3 consolidates the switch handling and removes channels.c from all tranmitters

Part 4 will address the emulator


src/target/drivers/input/switch/switch.c
```diff
@@ -0,0 +1,77 @@
...
+#include "target/drivers/mcu/stm32/rcc.h"
+
+extern s32 ADDON_ReadRawInput(int channel);
+
+#define TO_PIN(...) (struct mcu_pin) { __VA_ARGS__ }
+
+#ifndef STOCK_INPUTS
+    #ifdef SWITCH_STOCK
+        #define STOCK_INPUTS SWITCH_STOCK
+    #else
```
PhracturedBlue: Lint Error:
```
27:  Line ends in whitespace.  Consider deleting these extra spaces.  
```

PR #830, opened 2019-03-21
[WIP] Support configuration of TBS Crossfire devices

hexfet:
This PR is for review and not intended for merging.  The three commits can be separate PRs.  Commit da99 improves half-duplex serial.  Commit a116 adds the ability for a protocol to have a custom configuration page. Commit 4904 implements the CRSF configuration pages. The last includes some data files for emulator builds that would not be in the final PR. The emulator build shows TX and RX devices and allows navigating through the parameters, but commands are not emulated.

The CRSF configuration system works by message exchange between the transmitter (deviation) and the TBS devices.  This implementation has been tested with the full-size Crossfire transmitter and Nano RX receiver.  Each TBS component supplies parameters defining its configuration options and menu structure.  See the protocol/crsf_params_tx._c and ..._rx._c emulator data files for examples.

The CRSF configuration page is started by clicking on the protocol name.  One step of the page init function is to send a TBS message to discover available devices.  The device names are then shown in a scrollable (currently limited to four entries maximum). The device info response includes the number of parameters associated with the device.  Clicking on a device name starts the crsfdevice page, which starts a sequence of messages to retrieve all the device parameters.  These are displayed according to the menu structure defined in the parameter information by TBS.  When a setting is changed a "set parameter" message is sent to the TBS device, which responds with a parameter info message that includes any changes made.  Parameters of type COMMAND open a dialog to show the command status.  Some commands require an Ok/Cancel response from the user before proceeding (but whether a command needs this check isn't known until the command is started...).

The TBS menu information takes considerable RAM space (e.g. 2010 bytes for Nano RX).  This implementation does not attempt any compression of the menu info as that will add considerable messiness. A new target definition HAS_CRSF_CONFIG controls whether the code is included in a build. I suggest (eventually) merging this feature but with HAS_CRSF_CONFIG false for all transmitters.  A future custom build system will then allow users to include it themselves. This commit has CRSF configuration enabled for Devo10, t8sg_v2, and t8sg_v2_plus.

Still a couple rough edges.  The TBS devices sometimes hide or un-hide menu items when other settings are changed.  The number of rows in the current scrollable are not updated to reflect these changes.  Some parameter names are too long to be fully displayed on the same line as the parameter value.

Video with navigation through some of the menus [is here.](https://youtu.be/jRHjCW1BbEQ)

Appreciate all thoughts and comments.  Is there a better location for crsf.h? (and better name?)

Following is my current to-do list for CRSF configuration feature.
1) ~~Add configuration pages for color displays (some page functions may move to the common files)~~ 
170d950 
2) ~~Support changing number and string parameters (none are used for TX and Nano RX).~~  b863acf1
3) ~~Adjust parameter value horizontal size based on max number of characters needed. Consider using two lines if necessary.~~ d655782 
4) ~~Currently dialogs are created in the parameter receive serial rx ISR because TBS doesn't indicate what type of dialog is needed until after the request is started. If this is not acceptable try setting a flag when dialog needed and create it in the page event function.~~ cb4773c 
5) ~~Update GUI as needed if menu items become hidden or un-hidden.~~ 4d333bba, efe443ed
6) ~~Show indicator during initial retrieval of parameters from device.~~ 52537355


src/pages/common/_crsfdevice_page.c
```diff
@@ -0,0 +1,489 @@
...
+static char *recv_param_ptr;
+
+#define SEND_MSG_BUF_SIZE  16
+static u8 send_msg_buffer[SEND_MSG_BUF_SIZE];
+static volatile int send_msg_buf_count;     // tx data available semaphore with CRSF protocol
+
+#define CRSF_MAX_STRING_BYTES  2200     // max observed is 2010 in Nano RX
+#define CRSF_STRING_BYTES_AVAIL(current)  (CRSF_MAX_STRING_BYTES-((char *)(current)-strings))
+static char strings[CRSF_MAX_STRING_BYTES];
+static char *next_string = strings;
```
TheRealMoeder: This is a huge string. Can we move this to pagemem to save memory?

vladislavy:
t8sg_v2 "HAS_BUTTON_POWER_ON" is not used anymore. Also, it can be found at opentx\t12\hardware.h.

TheRealMoeder:
also, HAS_CRSF_CONFIG should be named SUPPORT_CRSF_CONFIG

hexfet:
Then does the default value for SUPPORT_CRSF_CONFIG go in target_all.h?

src/pages/common/_crsfdevice_page.c
hexfet: Moving to pagemem saves 570 bytes.  I'll push the change after testing.

TheRealMoeder: Maybe other variables which logically only belong to the crsf page could move to pagemem as well. But since the ram save is "only" 570 bytes we wouldn't gain anything, because pagemem union is already dictated by char strings[CRSF_MAX_STRING_BYTES]

src/pages/common/_crsfconfig_page.c
```diff
@@ -0,0 +1,83 @@
...
+};
+#endif
+
+static struct crsfconfig_page * const mp = &pagemem.u.crsfconfig_page;
+static struct crsfconfig_obj * const gui = &gui_objs.u.crsfconfig;
+static u16 current_selected = 0;
+
+static u32 last_update;
+static u8 number_of_devices;    // total known
+
```
TheRealMoeder: last_update and number_of_devices can probably also move to pagemem

src/pages/common/_crsfconfig_page.c
```diff
@@ -0,0 +1,83 @@
...
+ You should have received a copy of the GNU General Public License
+ along with Deviation.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#if HAS_CRSF_CONFIG
+
+#include "crsf.h"
+
+#ifndef EMULATOR
+crsf_device_t crsf_devices[CRSF_MAX_DEVICES];
```
TheRealMoeder: this could also move to pagemem

src/pages/128x64x1/crsfdevice_page.c
```diff
@@ -0,0 +1,221 @@
...
+    case UINT8:
+    case UINT16:
+    case FLOAT:
+    case INT8:
+    case INT16:
+    case STRING:
+    case INFO:
+    case OUT_OF_RANGE:
+    default:
+        break;
```
TheRealMoeder: will all the other cases get filled? If not, please make sure the compiler optimizes this to a single if statement

src/pages/128x64x1/crsfdevice_page.c
```diff
@@ -0,0 +1,221 @@
...
+            EDIT_LABEL_WIDTH, LINE_HEIGHT, &LABEL_FONT,
+            crsf_name_cb, NULL, (void *)param);
+        GUI_CreateLabelBox(&gui->value[relrow].lbl, EDIT_VALUE_X, y,
+            EDIT_VALUE_WIDTH, LINE_HEIGHT, &LABEL_FONT,
+            crsf_value_cb, NULL, (void *)param);
+        break;
+    case FOLDER:
+        GUI_CreateLabelBox(&gui->name[relrow], LABEL_X, y,
+            LABEL_WIDTH, LINE_HEIGHT, &LABEL_FONT, crsf_name_cb, press_cb, (void *)param);
+        break;
```
TheRealMoeder: Is this really different to default case?

TheRealMoeder:
> Then does the default value for SUPPORT_CRSF_CONFIG go in target_all.h?

Good question @howard0su @PhracturedBlue

src/pages/common/_crsfdevice_page.c
hexfet: Yes, it's just a lot of string space needed for all the parameters.  I've emailed TBS to see if they'll share information on the maximum space needed by each device.

For the transmitters where it's likely to be used there's still lots of space left :)

src/pages/128x64x1/crsfdevice_page.c
hexfet: This is resolved in b863acf1 by eliminating the switch since only one case was used.

src/pages/128x64x1/crsfdevice_page.c
hexfet: The default was eliminated in b863acf1 by handling all the real case values.

hexfet:
Commit 66def479 changes the feature define to SUPPORT_CRSF_CONFIG.  Default set to 0 in target_all.h.

src/pages/common/_crsfconfig_page.c
hexfet: Only last_update, as number_of_devices is need for smooth "pop" to config page from device page.  Commit ee3c2ab4

hexfet:
> 
> 
> Can we translate string?

The strings are supplied by the TBS devices so translation would require some special steps.  I've captured (most) of the strings that come from the full-size tx and Nano rx, but don't have any from other devices.  I've emailed TBS to ask for all possibilities.  The translated strings could be stored and then referenced when parameters are loaded from a device.

The device info includes a "parameters version" number.  If TBS maintains it then it could be used to detect if the current translations match the parameters in use.

TheRealMoeder:
> > Can we translate string?
> 

I meant the „Loading“

hexfet:
> 
> 
> > > Can we translate string?
> 
> I meant the „Loading“

Ah okay :) That should work.

hexfet:
Thanks for the comments.  Closing this PR to open a clean one without emulator data files.

PR #839, opened 2019-03-31
Fix oled display problem where display remained dark

hexfet:
Code refactoring missed a constant change that affected SPI clock speed to display.


PhracturedBlue:
What am I missing, the comment should still apply, unshifting OLED_SPI_RATE seems appropriate

hexfet:
OLED_SPI_RATE is defined as SPI_CR1_BR_FPCLK_DIV_8 which is already the un-shifted value.  That's what's expected by spi_set_baudrate_prescaler.  The value in the declaration initialization is shifted when spi_init_master is called in _spi_init.

PR #842, opened 2019-04-02
Fix 3x2 switch configu for devo7e

### LINKED ISSUE #838: Devo 7e 3x2 switches not working
With current version nightly & build from master the 3x2 switches are not working properly. Both switches SWA & SWB read wrong values:
0 position gets 0
1 position gets 0
2 position gets 1
Going back to Nov 2017 versions gets it working fine again. Details posted in:
[https://www.deviationtx.com/forum/6-general-discussions/8361-devo-7e-3x2-switches-not-working-anymore#73614](url)

PhracturedBlue:
Fixes #838


src/target/tx/devo/devo7e/Makefile.inc
```diff
@@ -1,13 +1,19 @@
 SCREENSIZE  := 128x64x1
 DISPLAY_DRIVER := spi/128x64x1.c
 FILESYSTEMS := common base_fonts 128x64x1
+ifdef NEW_BOOTLOADER
```
TheRealMoeder: Not related to the fix, but I'm fine with keeping it.

src/target/tx/devo/devo7e/hardware.h
```diff
@@ -29,20 +29,24 @@
...
     EXTRA_2WAY(INP_SWA, 0x04, 0x08, CHAN_INVERT, SWITCH_2x2) \
     EXTRA_2WAY(INP_SWB, 0x01, 0x02, CHAN_INVERT, SWITCH_2x2)
 
-#define SWITCH_2x2  ((1 << INP_SWA0) | (1 << INP_SWA1) | (1 << INP_SWB0) | (1 << INP_SWB1))
-#define SWITCH_3x1  ((1 << INP_SWB0) | (1 << INP_SWB1) | (1 << INP_SWB2))
-#define SWITCH_3x2  ((1 << INP_SWA0) | (1 << INP_SWA1) | (1 << INP_SWA2) | SWITCH_3x1)
-#define SWITCH_STOCK (~(SWITCH_2x2 | SWITCH_3x2))
+#define ADDON_SWITCH_CFG \
+    ADDON_SWITCH("2x1", SWITCH_2x2, 0) \
+    ADDON_SWITCH("2x2", SWITCH_2x1, 0) \
```
hexfet: Don't have a way to test this, but asking because it looks odd.  Is it right that the string doesn't match the constant name in these two lines?

PhracturedBlue: no....I don't have a 2x2 to test on either.  Good catch though

src/target/tx/devo/devo7e/hardware.h
```diff
@@ -29,20 +29,24 @@
...
-    ADDON_SWITCH("2x2", SWITCH_2x2, 0) \
-    ADDON_SWITCH("3x1", SWITCH_3x1, 0) \
-    ADDON_SWITCH("3x2", SWITCH_3x2, 0)
-
 #define EXTRA_SWITCHES \
-    EXTRA_3WAY(INP_SWA, 0x04, 0x08, CHAN_NONINV, SWITCH_3x2) \
+    EXTRA_3WAY(INP_SWA, 0x04, 0x08, CHAN_NONINV, SWITCH_3x1) \
     EXTRA_3WAY(INP_SWB, 0x01, 0x02, CHAN_NONINV, SWITCH_3x2) \
     EXTRA_2WAY(INP_SWA, 0x04, 0x08, CHAN_INVERT, SWITCH_2x2) \
     EXTRA_2WAY(INP_SWB, 0x01, 0x02, CHAN_INVERT, SWITCH_2x2)
```
hexfet: This seems odd as well because it's no longer symmetric.  Not sure what's correct.

PhracturedBlue: the 2way switch is also broken here.  I'm not sure we actually support a 2x1 configuration actually.   Previously we did not.  I was trying to more closely align with how the 7e-256 is setup.

hexfet: Makes sense.

PR #844, opened 2019-04-02
[E016H protocol] Add calibration flag

goebish:
Add calibration flag for E016H format.
Works same as stock controller, set both sticks to external bottom corners to start accelerometer calibration.


hexfet:
Would it be better to use a channel to control the calibration flag?  Then users can set up a mixer for either stick activation or use a button or switch.

goebish:
Most of those small quads can be calibrated like that, but only a few ones send a flag, I don't think that's worth using an extra channel.

PR #846, opened 2019-04-06
Add tool to scan for and dump XN297 packets

TheRealMoeder:
With the help of @goebish I added a feature to scan for valid XN297 packets and dump their payload. This will help in identifying minor protocol modifications (e.g. new flags) and deviating protocols which run on integrated MCU+XN297 controllers.

To Do:
- [x] Add text GUI
- [x] ~~Move scanner and XN297Dump together into a new RF Tools submenu~~ should do this in seperate PR
- [x] remove xn297dump from default devo7e build
- [x] ~~use common scramble table for CRC check~~ -> goebish will do that
- [x] ~~Move scanner and xn297dump memory into union to save memory~~ not really worth the saving
- [x] Add packet interval measurement (accuracy should be within 10-20 µs for most protocols if averaging is long enough
- [x] Test modular build -> modular build works fine on t8sg_v2_plus and even fits current devo7e if enabled (weighs 1.47 kB)
- [x] Rework dumper to enable unscrambled mode and other features
- [x] Allow dumping to datalog.bin


TheRealMoeder:
@PhracturedBlue can you test if color gui is usable on a real tx? Not sure with the screen update rates...

src/protocol/xn297dump_nrf24l01.c
```diff
@@ -0,0 +1,294 @@
...
+#define MAX_PACKET_LEN     32
+#define CRC_LENGTH         2
+#define MAX_RF_CHANNEL     84
+#define DUMP_RETRIES       10  // stay on channels long enough to capture packets
+#define INTERVAL_AVERAGE   4
+
+static const char *const xn297dump_opts[] = {
+    _tr_noop("Bitrate"), "1 Mbps", "2 Mbps", "250 Kbps", NULL,
+    _tr_noop("Address"), "5 byte", "4 byte", "3 byte", NULL,
+    _tr_noop("Scrambled"), _tr_noop("No"), _tr_noop("Yes"), NULL,
```
goebish: I'd set Scrambled to Yes by default as this is the most used mode

TheRealMoeder: That's what I thought too, but XN297_SCRAMBLED is 1 in iface_nrf24l01.h, and I didn't like the idea of defining it reversed within the protocol. If I rename the protopt "Unscrambled it might be  even more confusing. Suggestion?

hexfet: My two cents is that the values don't need to match.  The default option Yes is a presentation "thing" that means the code does whatever it needs to enable scrambling.

goebish: No big deal, my changes for xn297@250kbps with cc2500 will require to explicitly state scrambling, I'll swap the on/off value if required then.
https://github.com/DeviationTX/deviation/pull/860/commits/78e85a24e18a9376d2f819e85c2c1c6e8cb7fdc8

PR #849, opened 2019-04-09
Updates for S.Port telemetry

hexfet:
- Add support for S.Port sensors ACCX, ACCY, ACCZ, and AIR_SPEED
- Add support for Qlrs sensors PITCH and ROLL (displayed in same screen position as ACCX and ACCY)
- Add telemetry test page value reset (long press UP) for PXX protocol
- Clean up telemetry value reset code


TheRealMoeder:
200 byte RAM increase for devo7e, probably to the large gui_label array which is increased by 4, although not needed in non HAS_EXTENDED_TELEMETRY

PR #861, opened 2019-04-19
Fix t8sg V1 and V2 LCD contrast

vladislavy:


src/target/drivers/display/spi/128x64x1_nt7538.h
```diff
@@ -49,4 +49,6 @@ inline static void _lcd_init()
     LCD_Cmd(0x2F);      // Power Controller: VFollower ON
     i = 0x8000;
     while (i) i--;
+    if (HAS_LCD_FLIPPED)
+        LCD_Cmd(0x26);  //Select Internal Resistor Rate (Rb/Ra)
```
PhracturedBlue: Lint Error:
```
53:  Should have a space between // and comment  
```

PR #871, opened 2019-05-15
Protocol e010 cc2500

goebish:
Add a new protocol for EAchine E010 using the CC2500 instead of NRF24L01

Tested.


.gitignore
```diff
@@ -25,6 +25,7 @@ utils/genspeech/*.zip
 #src/beta-*
 /local/
 coverage/
+src/.vs/
```
TheRealMoeder: do we want this?

goebish: I want it and I suppose that won't disturb anyone

PR #882, opened 2019-06-10
Remove standard GUI from Devo 6 and 8

hexfet:
This fixes the issue with current master failing to build due to rom space overrun.  The space is exceeded after merging PR #880 .  Pending PR #881 makes the same change only for Devo 8.


vladislavy:
Disable HAS_4IN1_FLASH can save about 800 bytes of rom space.

hexfet:
800 bytes is enough for the current PRs to build successfully.  Is that a better solution than removing standard GUI?  Will any users be affected by disabling 4IN1 for the 6 and 8?

vladislavy:
4IN1_FLASH has sense for transmitters with small flash memory (Devo F4, F7, Radiolink AT9). Also, iRangeX IR8M use 4IN1_FLASH code for managing ordinary 4in1 module. Any other transmitter will be better to use with ordinary (devo) 4in1 module without code for 4IN1_FLASH.

hexfet:
Just the 4IN1_FLASH change is no longer enough to build the pending pull requests.  It is a good idea though so updated this PR to make that change.

vladislavy:
Probably PR can be closed since now we have spare ROM space.

PR #950, opened 2019-12-30
Please increase the maximum of the custom sounds

eried:


src/music.h
```diff
@@ -29,7 +29,7 @@ enum Music {
 };
 
 #if HAS_EXTENDED_AUDIO
-#define MAX_VOICEMAP_ENTRIES 400 // arbitraty chosen
+#define MAX_VOICEMAP_ENTRIES 500 // arbitraty chosen
```
```
32:  At least two spaces is best between code and comments  
```

```
32:  At least two spaces is best between code and comments  
```

eried:
Thanks! I added some sound fx and noticed that after the internal id 217 my controller was hanging. Limit can be lower, 450 or so, just to have space for around 60-100 extra custom sounds

https://github.com/eried/Research/tree/master/JumperT8SG/soundfx

TheRealMoeder:
Development around deviation is rather quiet, we might as well merge this. If we ever run into ram issues, we can rethink things.

Issue #960, opened 2020-02-16
CUSTOM SOUNDS should be skippable, and not every sound should go to the queue


eried:
Problem: The voice/sound system is too rudimentary.

1) Sounds are never skipped: Welcome sound, and custom sounds should be skipped if another sound is triggered. Except if the sound is an alarm, or part of the alarm (like the time of that alarm).

2) Every sound goes to a queue: Only one item should be in the queue for custom sounds. Imagine you quickly move a switch with a sound assigned N times. What is the point of hearing all the N times VS only the last state of the switch, which is the valid one?

3) Custom sounds should not be triggered immediately: If you have a switch with 3 sounds and you switch from A to C. Why should you hear B and then C, not only C?

somewhatlurker:
It's nice to have something I can test in emulator actually, unlike USBHID, so let's brainstorm here a bit.

The currently proposed solution seems to treat all custom alarms (everything configured in the voice config menu?) as equal and like they should be replaced when any other custom alarm triggers.  
And due to the potential for this to unintentionally mask alerts from other switches, this is deemed unsatisfactory.

Proposed solutions to that new behaviour so far are either to assign audio to groups which can be skipped by other triggers in that group (one switch = one group, or perhaps leave it user-configurable), or to give alerts a priority setting where only equal or lower priority alerts will be replaced (so critical alerts such as one for an arm/disarm switch will always be played, but less important alerts will be skipped by other actions).

Is this summary correct?

　

I can see the appeal of both approaches -- a pilot with alerts setup on switches probably does want the changes to be acknowledged even after making other changes, but when there's constant alerts for several seconds after changing two or three switches a pilot may start to zone out and not notice something more important being played.  
The concern over pilots zoning out could be potentially mitigated by offering ready-to-go voice packs at multiple speeds without code changes though.

eried:
Yes that is a good summary. I do not actually agree on the importance of the alarms though, if I start hearing an alarm and switch flight mode from LEVEL -> HORI -> ACRO, I do not really expect to hear the whole alarm sounds sequences, then HORI, then ACRO like it is now. I would expect the alarm to be abrupted with only "ACRO" voice.

somewhatlurker:
And I don't really know if I like the idea of interrupting alerts for unrelated things when inputs are received...
[...]

PR #961, opened 2020-02-16
New method for handling custom alarms

eried:
This fixes https://github.com/DeviationTX/deviation/issues/960

Available settings in `src/music.c`:
`USE_NEW_CUSTOM_ALARM_MODE`, if is `0` then the behaviour is exactly the same as the original

`CUSTOM_ALARM_IGNORE_MS`, is the time in ms that the sound will not be triggered. So if is 200 ms and the user switch from A to C, sound B wont be played if the user took less than 200 ms on moving the switch from B to C.

### DIFF
--- a/src/extended_audio.c
+++ b/src/extended_audio.c
@@ -248,8 +248,29 @@ int AUDIO_AddQueue(u16 music) {
         printf("Voice: mp3 length is zero\n");
         return 0;
     }
-
-    audio_queue[num_audio++] = music;
+    
+    int new_num_audio = -1;
+    
+#if USE_NEW_CUSTOM_ALARM_MODE
+    u32 t = CLOCK_getms();
+    
+    // Replace the last custom alarm id, if there is already one
+    if(voice_map[music].id>=CUSTOM_ALARM_ID)
+    {
+        u32 id = voice_map[audio_queue[next_audio-1]].id;
+        if(id==0 || id>=CUSTOM_ALARM_ID) // Only skip welcome message or custom alarms
+            audio_queue_time = t + CUSTOM_ALARM_IGNORE_MS; // Do not consume the audio right away (prevent triggering transitional sounds)
+        
+        for(u8 i=next_audio; i<num_audio; i++)
+            if(voice_map[audio_queue[i]].id>=CUSTOM_ALARM_ID)
+                new_num_audio = i;
+    }
+#endif
+        
+    if(new_num_audio < 0)
+        audio_queue[num_audio++] = music; // Add to queue
+    else
+        audio_queue[new_num_audio] = music; // Replace queue item, do not enqueue custom sounds
     return 1;
 }
 
--- a/src/music.h
+++ b/src/music.h
@@ -37,6 +37,8 @@ enum Music {
 #endif
 #define VOICE_UNIT_OFFSET 130
 #define CUSTOM_ALARM_ID 200 // start of custom MP3 IDs
+#define USE_NEW_CUSTOM_ALARM_MODE 1 // if this is 0, all the custom alarms are enqueued and never skipped
+#define CUSTOM_ALARM_IGNORE_MS 200 // ignore custom alarms if the switch was faster than this value (requires USE_NEW_CUSTOM_ALARM_MODE)
 #define VOICE_DEC_SEP 110  // MP3 ID of DECSEP = 110 + MUSIC_TOTAL
 #define NUM_STICKS 4
 #define NUM_AUX_KNOBS	(INP_HAS_CALIBRATION - NUM_STICKS)	// Exclude sticks


src/music.h
```diff
@@ -37,6 +37,8 @@ enum Music {
 #endif
 #define VOICE_UNIT_OFFSET 130
 #define CUSTOM_ALARM_ID 200 // start of custom MP3 IDs
+#define USE_NEW_CUSTOM_ALARM_MODE 1 // if this is 0, all the custom alarms are enqueued and never skipped
+#define CUSTOM_ALARM_IGNORE_MS 200 // ignore custom alarms if the switch was faster than this value (requires USE_NEW_CUSTOM_ALARM_MODE)
```
```
40:  At least two spaces is best between code and comments  
41:  At least two spaces is best between code and comments  
```

src/extended_audio.c
```diff
@@ -248,8 +248,29 @@ int AUDIO_AddQueue(u16 music) {
         printf("Voice: mp3 length is zero\n");
         return 0;
     }
-
-    audio_queue[num_audio++] = music;
+    
+    int new_num_audio = -1;
+
```
PhracturedBlue: Lint Error:
```
251:  Line ends in whitespace.  Consider deleting these extra spaces.  
253:  Line ends in whitespace.  Consider deleting these extra spaces.  
```

src/extended_audio.c
```diff
@@ -248,8 +248,29 @@ int AUDIO_AddQueue(u16 music) {
...
-
-    audio_queue[num_audio++] = music;
+    
+    int new_num_audio = -1;
+    
+#if USE_NEW_CUSTOM_ALARM_MODE
+    u32 t = CLOCK_getms();
+    
+    // Replace the last custom alarm id, if there is already one
+    if(voice_map[music].id>=CUSTOM_ALARM_ID)
```
PhracturedBlue: Lint Error:
```
256:  Line ends in whitespace.  Consider deleting these extra spaces.  
258:  Missing spaces around >=  
258:  Missing space before ( in if(  
```

src/extended_audio.c
```diff
@@ -248,8 +248,29 @@ int AUDIO_AddQueue(u16 music) {
...
+    {
+        u32 id = voice_map[audio_queue[next_audio-1]].id;
+        if(id==0 || id>=CUSTOM_ALARM_ID) // Only skip welcome message or custom alarms
+            audio_queue_time = t + CUSTOM_ALARM_IGNORE_MS; // Do not consume the audio right away (prevent triggering transitional sounds)
+        
+        for(u8 i=next_audio; i<num_audio; i++)
+            if(voice_map[audio_queue[i]].id>=CUSTOM_ALARM_ID)
```
PhracturedBlue: Lint Error:
```
261:  At least two spaces is best between code and comments  
261:  Missing spaces around ==  
261:  Missing space before ( in if(  
262:  At least two spaces is best between code and comments  
263:  Line ends in whitespace.  Consider deleting these extra spaces.  
264:  Missing spaces around =  
264:  Missing spaces around <  
264:  Missing space before ( in for(  
265:  Missing spaces around >=  
265:  Missing space before ( in if(  
```

src/extended_audio.c
```diff
@@ -248,8 +248,29 @@ int AUDIO_AddQueue(u16 music) {
...
+                new_num_audio = i;
+    }
+#endif
+        
+    if(new_num_audio < 0)
+        audio_queue[num_audio++] = music; // Add to queue
+    else
+        audio_queue[new_num_audio] = music; // Replace queue item, do not enqueue custom sounds
```
PhracturedBlue: Lint Error:
```
269:  Line ends in whitespace.  Consider deleting these extra spaces.  
270:  Missing space before ( in if(  
271:  At least two spaces is best between code and comments  
273:  At least two spaces is best between code and comments  
```

eried:
> I like the general idea, but I'd suggest working this into the config (tx.ini?) files instead of the define and also add menu options for setting this up? Maybe we can have the "old" behavior when the value for CUSTOM_ALARM_IGNORE_MS is "0"?

I agree tha a menu option would be desirable, I am going to check that section. I left that as a define thinking that in the unlikely case that someone might want that old behaviour, it still could be engaged.

> 
> Edit:
> Another thing comes to my mind: Maybe ignoring all custom alarms coming in fast isn't desirable: Imaging flipping a switch immediately before a telemetry alarm with custom sound comes in. So maybe this behavior should be exclusive for switch custom alarms? I'm open for discussion on this matter.

You are right, I was not aware or explored those telemetry alarms and others. In this case it would require a "source" flag (i.e. priority and have that as a setting in the menu maybe?).

I am thinking a solution could be:
1) Add priority to the `int AUDIO_AddQueue(u16 music) {` function
2) Expose that priority in options:
**Priority for switch voice: high|normal**

In that way the "buzzer" functionality (that was what I wanted to fix later) could use those rules.

I was just blinded for how cool it feels that the sounds are skipabble. https://youtu.be/Bhd3Ko32Q8U?t=264

TheRealMoeder:
> I agree that a menu entry would be ideal, I am going to check that section on the source. I left that as a `define` thinking that in the unlikely case that someone might want that old behaviour, it still could be engaged. 

Menu is pretty simple; I can add that later in case you don't want to dig into the source too much.

> You are right, I was not aware or explored those telemetry alarms and others. In this case it would require a "source" flag (i.e. priority and have that as a setting in the menu maybe?).
> I am thinking a solution could be:
> 1. Add priority to the `int AUDIO_AddQueue(u16 music) {` function
> 2. Expose that priority in options:
>    **Priority for switch voice: high|normal**

I think this really only applies to switches, so we might as well make a seperate case for switches. How about somehow passing a switch id (or group) together with the voiceid for the queue voice to always replace queue entries for the same switch id/group so we only have the latest switch state in the queue? Will probably require a seperate AUDIO_AddQueueSwitch (u16 music u16 group) or similar. The reason behind this is, you could also be flipping to switches at the same time (not that uncommon in the heat of the flight) and wouldn't want to lose any of the announcements of either switch.

While at reworking, I think it also makes sense for AUDIO_AddQueue to check, whether the music being added to the queue is the same as the last entry in the queue, and in that case discard the duplicate entry. 

 
> In that way the "buzzer" functionality (that was what I wanted to fix later) could use those rules.
> 
> I was just blinded for how cool it feels that the sounds are skipabble. https://youtu.be/Bhd3Ko32Q8U?t=264

I wasn't aware that someone finally found a fix for the "saving" issue - great!

@goebish you always have some good ideas, any comments on this issue?

eried:
> Menu is pretty simple; I can add that later in case you don't want to dig into the source too much.

That would be awesome.

> While at reworking, I think it also makes sense for AUDIO_AddQueue to check, whether the music being added to the queue is the same as the last entry in the queue, and in that case discard the duplicate entry.

Not only duplicates, but also how many ms since it was triggered, to skip the sounds while switching a 3 state switch. I do not really agree with keeping ALL the switches voices all the time, I like how it drops or interrupts voices. But ofc this would need some kind of survey, or option.

> I wasn't aware that someone finally found a fix for the "saving" issue - great!

Yes, sadly is a hardware issue that will need a pcb revision.

TheRealMoeder:
> > While at reworking, I think it also makes sense for AUDIO_AddQueue to check, whether the music being added to the queue is the same as the last entry in the queue, and in that case discard the duplicate entry.
> 
> Not only duplicates, but also how many ms since it was triggered, to skip the sounds while switching a 3 state switch. I do not really agree with keeping ALL the switches voices all the time, I like how it drops or interrupts voices. But ofc this would need some kind of survey, or option.

Well, the switch issue would be solved with adding the alarm group to a queue entry. Which means an alarm from the same group will cancel old alarms in the queue with the same group id. At least that's how thought of it. Do you get what I mean?

eried:
Yes I understand about the "groups", that concept makes skipping voices harder to implement, and it just solves the "switching from A to C without hearing B". 

Just for the sake of it, I am going to list the cases:
1) Able to **skip** currently playing voices:
[video](https://www.youtube.com/watch?v=sRWv9bmFc68)

2) **Except** if is an alarm or some important stuff:
[video](https://www.youtube.com/watch?v=uDyiwnTtNr8)

Those two are common sense, and not open for discussion, right?
3) **Skip** voices from the _same switch_:
[video](https://www.youtube.com/watch?v=IkDMQHpxRYs)

This is also common sense. However:
4) **Skip** pending voices, even if they are _not from the same switch_:
[video](https://www.youtube.com/watch?v=JqpUHWiBZno)

In my opinion, in that example the user should not have to hear all the switches positions in a queue. That could be the setting of "priority of switch position", however I still do not see much of a point of enqueueing (and hearing) all voice switches.

TheRealMoeder:
> 1. Able to **skip** currently playing voices:
> 2. **Except** if is an alarm or some important stuff:
> Those two are common sense, and not open for discussion, right?

Right, would be solvable with the switch group concept.

> 3) **Skip** voices from the _same switch_:

This was what started the whole discussion, so nothing to add here.

> 4) **Skip** pending voices, even if they are _not from the same switch_:
> In my opinion, in that example the user should not have to hear all the switches positions in a queue. That could be the setting of "priority of switch position", however I still do not see much of a point of enqueueing (and hearing) all voice switches.

This is the only case which would need the priority setting. But I'm having some issues about making setting each alarms priority a requirement. From a user perspective, I think telemetry, timer and battery alarms are all somehow high priority unless specifically changing this behavior, so this will probably be the default behavior. Your wish number 4 could be achievable by having the option to "upgrade" specific switch settings to the same high priority level as the ones above.

I do not think we can omit enqueuing all switches, as this will make the behavior somewhat unpredictable and I again think of the case of dual switch flips at the same time.

eried:
I am still thinking on a solution, so question for you @TheRealMoeder, Imagine 2 switches, you switch one, voice start playing... switch the 2nd one. Does the first voice gets interrupted?

TheRealMoeder:
> I am still thinking on a solution, so question for you @TheRealMoeder, Imagine 2 switches, you switch one, voice start playing... switch the 2nd one. Does the first voice gets interrupted?

I can imagine both behaviors, but I tend to prefer the first voice not being interrupted

src/protocol/usbhid.c
```diff
@@ -63,7 +63,7 @@ static u16 usbhid_cb()
         
     HID_Write(packet, sizeof(packet));
 
-    return 50000;
+    return 400; // faster refresh
```
PhracturedBlue, line 66: Lint Error:
```
66:  At least two spaces is best between code and comments  
```

src/pages/128x64x1/advanced/mixer_curves.c
```diff
@@ -89,15 +90,16 @@ void PAGE_EditCurvesInit(int page)
...
 }
 
 static unsigned action_cb(u32 button, unsigned flags, void *data)
 {
     (void)data;
-    if ((flags & BUTTON_PRESS) || (flags & BUTTON_LONGPRESS)) {
+    if ((flags & BUTTON_PRESS) || (flags & BUTTON_LONGPRESS) || (flags & BUTTON_RELEASE)) {
         if (CHAN_ButtonIsPressed(button, BUT_EXIT)) {
-            PAGE_Pop();
+            if (flags & BUTTON_RELEASE)
```
PhracturedBlue, line 101: Lint Error:
```
101:  Line ends in whitespace.  Consider deleting these extra spaces.  
```

PR #964, opened 2020-03-01
Add power on delay for t8sg_v2 series

vladislavy:


src/main.c
```diff
@@ -47,7 +48,21 @@ int main() {
 #ifndef ENABLE_MODULAR
     //Banner();
 #endif
+
+#if defined(HAS_BUTTON_POWER_ON) && HAS_BUTTON_POWER_ON
+    u32 power_on_delay = 0;
+    while (1) {
+        if(!gpio_get(_PWRSW_PORT, _PWRSW_PIN))
+            PWR_Shutdown();
+        else {
```
PhracturedBlue: Lint Error:
```
55:  Missing space before ( in if(  
57:  If an else has a brace on one side, it should have it on both  
```

src/main.c
```diff
@@ -47,7 +48,21 @@ int main() {
...
+            power_on_delay++;
+            if (power_on_delay > 500000)  // about 1 sec
+                break;
+        }
```
PhracturedBlue: Lint Error:
```
61:  Line ends in whitespace.  Consider deleting these extra spaces.  
```

PR #965, opened 2020-03-01
Add power on delay to t8sg_v2 series

vladislavy:


src/main.c
```diff
@@ -47,7 +48,21 @@ int main() {
 #ifndef ENABLE_MODULAR
     //Banner();
 #endif
+#if defined(HAS_BUTTON_POWER_ON) && HAS_BUTTON_POWER_ON
+    u32 power_on_delay = 0;
+    while (1) {
+        if (!gpio_get(_PWRSW_PORT, _PWRSW_PIN)) {
+            PWR_Shutdown();
+        }
+        else {
```
PhracturedBlue: Lint Error:
```
57:  If an else has a brace on one side, it should have it on both  
```

src/main.c
```diff
@@ -47,7 +47,22 @@ int main() {
 #ifndef ENABLE_MODULAR
     //Banner();
 #endif
+#if defined(HAS_BUTTON_POWER_ON) && HAS_BUTTON_POWER_ON
+    u32 power_on_delay = 0;
+    while (1) {
+        u32 value = GPIO_pin_get(PWR_SWITCH_PIN);
+	if ((HAS_PWR_SWITCH_INVERTED && value)
+	    || (!HAS_PWR_SWITCH_INVERTED && !value)) {
```
PhracturedBlue: Lint Error:
```
54:  Tab found; better to use spaces  
55:  Tab found; better to use spaces  
```

PR #972, opened 2020-03-26
Add "USB drive erase" procedure

vladislavy:
USB drive erase procedure
1) "Power ON" transmitter with USB mode (ENT + Power ON)
2) Press "UP" button
3) After short beep press "DOWN" button
4) After short beep "Power OFF" transmitter
5) "Power ON" transmitter with USB mode (ENT + Power ON)
6) Format transmitter USB drive with "Quick Format" option enabled


eried:
Maybe on the usb logo a small text could appear with those instructions? I could change the bitmap

vladislavy:
I'm not sure we need this procedure for all transmitters, but USB logo is common for every GUI size. Probably we need restrict USB drive erase procedure to T8SG series and IR8M.

PR #997, opened 2020-08-09
Advcurve ui bugfix

eried:
There are some small details in the interface of mixer -> curves editor

1. The titlebar is selectable but it is not in reality.
2. The curve value should be selected by default

image

### DIFF
--- a/src/pages/128x64x1/advanced/mixer_curves.c
+++ b/src/pages/128x64x1/advanced/mixer_curves.c
@@ -58,7 +58,8 @@ void PAGE_EditCurvesInit(int page)
     }
     edit->curve = *curve;
 
-    GUI_CreateTextSelectPlate(&gui->name, NAME_X, 0, NAME_W, HEADER_HEIGHT, &TEXTSEL_FONT, NULL, set_curvename_cb, NULL);
+    //GUI_CreateTextSelectPlate(&gui->name, NAME_X, 0, NAME_W, HEADER_HEIGHT, &TEXTSEL_FONT, NULL, set_curvename_cb, NULL);
+    GUI_CreateLabelBox(&gui->title, NAME_X, 0 , NAME_W, HEADER_HEIGHT, &LABEL_FONT, GUI_Localize, NULL, _tr(CURVE_GetName(tempstring, curve)));
     GUI_CreateButtonPlateText(&gui->save, SAVE_X, 0, SAVE_W, HEADER_WIDGET_HEIGHT, &BUTTON_FONT , GUI_Localize, okcancel_cb, _tr_noop("Save"));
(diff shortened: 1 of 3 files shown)


src/pages/128x64x1/guiobj.h
```diff
@@ -270,7 +270,8 @@ struct crsfdevice_obj {
 
 /****Advanced ****/
 struct advcurve_obj {
-    guiTextSelect_t name;
+    //guiTextSelect_t name;
```
PhracturedBlue, line 273: Lint Error:
```
273:  Should have a space between // and comment  
```

src/pages/common/advanced/_mixer_curves.c
```diff
@@ -17,7 +17,7 @@ static struct advcurve_obj * const gui  = &gui_objs.u.advcurve;
 static struct curve_edit   * const edit = &pagemem.u.mixer_page.edit;
 
 static void okcancel_cb(guiObject_t *obj, const void *data);
-static const char *set_curvename_cb(guiObject_t *obj, int dir, void *data);
+//static const char *set_curvename_cb(guiObject_t *obj, int dir, void *data);
```
PhracturedBlue, line 20: Lint Error:
```
20:  Should have a space between // and comment  
```

src/pages/128x64x1/advanced/mixer_curves.c
```diff
@@ -58,7 +58,8 @@ void PAGE_EditCurvesInit(int page)
     }
     edit->curve = *curve;
 
-    GUI_CreateTextSelectPlate(&gui->name, NAME_X, 0, NAME_W, HEADER_HEIGHT, &TEXTSEL_FONT, NULL, set_curvename_cb, NULL);
+    //GUI_CreateTextSelectPlate(&gui->name, NAME_X, 0, NAME_W, HEADER_HEIGHT, &TEXTSEL_FONT, NULL, set_curvename_cb, NULL);
```
PhracturedBlue, line 61: Lint Error:
```
61:  Should have a space between // and comment  
```

eried:
Oh I am noticing that my fix does not work for /320x240x16/. 

Can anyone with those remotes that if they go into "Advanced curves" of a mixer like the one shown in the image on top, they can scroll between curve types?? I can fix this also there but I just need to confirm that it is also a problem there

TheRealMoeder:
From testing in the emulator on color displays the curve type also cannot be changed with in the curves editor.

TheRealMoeder:
Actually there is some behavior to the title bar in curve editor for both GUIs , as can be seen in the code. When in a multipoint curve type, you can immediatly change the number of curve points (3 through 12).

Issue #981, opened 2020-04-18
USBHID update rate is really slow


AlessandroAU:
Coming from an OpenTX platform I found the USBHID to be unflyable. Feels like the stick update rate is really slow. I filmed some slow-mo footage on a 100hz monitor of me spinning a stick in a circle on both a QX7 and a T8SG to illustrate the difference. 
QX7:
![ezgif-6-ec96658041af](https://user-images.githubusercontent.com/22009829/79638909-8c468f00-81cb-11ea-8b43-a9ae23bde64a.gif)
Devo:
![ezgif-6-731c691ac529](https://user-images.githubusercontent.com/22009829/79638912-9072ac80-81cb-11ea-86e4-6bf21257b94c.gif)

eried:
It is noticeable without slow footage? I do not have another controller to test but it seems smooth on my t8sg v2 with devo

AlessandroAU:
Yes very noticeable in velocidrone, you can even hear the difference. It's what lead me to investigate.

AlessandroAU:
Sharp turns in the sim sounds like D-term oscillations on a real quad and swift stick deflections are really jerky. Obviously hard to convey in a video but here it is anyway. First half is T8SG v2 (with hall gimabls) and second half is QX7. https://www.youtube.com/watch?v=oLlId8_gTKM&feature=youtu.be

eried:
Oh I see. Can you record with slow motion the stick monitor screen and the physical stick moving together to full left and then full right? For both controllers? To calculate the lag with an objective target. I can take a look on the USBhid code

eried:
[deviation-t8sg_v2_plus-v5.0.0-32d5597.zip](https://github.com/DeviationTX/deviation/files/4497275/deviation-t8sg_v2_plus-v5.0.0-32d5597.zip)

Can you test this one in the video?

AlessandroAU:
Sorry no video yet. Latency feels a bit better (I might just be getting used to the controller) but update rate feels the same.

eried:
Can you test this version now:
[deviation-t8sg_v2_plus-v5.0.0-32d5597.zip](https://github.com/DeviationTX/deviation/files/4499676/deviation-t8sg_v2_plus-v5.0.0-32d5597.zip)

I did some checkings against a normal game controller, and the previous version was quite slow. However this one is on par with the commercial controller:
[...]

PR #998, opened 2020-09-09
make USB HID period configurable (and default is now 8ms)

TheRealMoeder:


src/protocol/usbhid.c
```diff
@@ -18,6 +18,17 @@
...
+static const char * const usbhid_opts[] = {
+  _tr_noop("Period (ms)"),  "1", "64", NULL,
+  NULL
+};
+enum {
+    PROTO_OPTS_PERIOD,
+    LAST_PROTO_OPT,
+};
+ctassert(LAST_PROTO_OPT <= NUM_PROTO_OPTS, too_many_protocol_opts);
+#define USBHID_FRAME_PERIOD_STD 8 // 8ms default period for 125Hz
```
PhracturedBlue, line 30: Lint Error:
```
30:  At least two spaces is best between code and comments  
```

goebish (approved):

somewhatlurker:
Fun. Still gotta fix test stubs then, and I have no clue what to do for 7e and f7.  
I guess I'll leave the latter to someone who knows what the hell they're doing. There's no rush because after all this isn't even tested yet anyway. (previous discussion in #981)

TheRealMoeder:
@goebish can you help out with the linker finding everything in modular build? I tried, but something seems messed up with my build environment :(

goebish:
I can take a look but I'll be back home in 10 days

TheRealMoeder:
I think I found the issue. Can't simply make `usb_endpoint_descriptor hid_endpoint` non-const

somewhatlurker:
seems to be linking properly; that's good

somewhatlurker:
so all this needs is testing that it works as expected then

TheRealMoeder:
I'll try to do so this weekend.

TheRealMoeder:
Couldn't resist. Works perfect down to 1 ms in non modular build (t8sg as my testing device here). I have no modular deco right now, I might try building a modular build for my t8sg for testing purposes.

TheRealMoeder:
I noticed some interesting behavior in lower update rates. For example, with 50 ms, the average ist indeed 50 ms, but it actually alternates between 35 and 65 ms. But since nobody wants these slow update rates it's probably not worth investigating. From 1 to 10 ms i noticed very exact update rates.

somewhatlurker:
I'm not really sure about how USB endpoints work, but I think that's related to how the OS polls the device.

For example, Windows seems to take the provided bInterval and round it down to a closest power of two (or always 32 if larger than 32).  
That does sound like the jitter you're seeing, and should be do different to the original behaviour which used a 32ms bInterval.

It should be possible to hide that by always requesting a lower polling rate or only letting the user choose power-of-two periods somehow, but I don't know if it's worth adapting to "features" of USB stacks that don't really harm operation.

eried:
> but it actually alternates between 35 and 65 ms

How are you measuring this?

> From 1 to 10 ms i noticed very exact update rates.
True, and this is similar to an 8bitdo controller I tested as a example to compare the "laggy" usb, so we are good with 8ms :D

TheRealMoeder:
I measured using Wireshark on MacOS, but honestly only exemplary on settings 1-10 and 50 ms. But I'm not deep into USB stuff, just wanted a proof of the changes working. Also tried these settings in Picasim  simulator though wine on the same system, which shows the same behavior as the Wireshark results suggested. Since 50 ms isn't usable at all, it really is not worth investing there. Maybe we should limit the protocol option to 32.

somewhatlurker:
Yeah, I'd agree that limiting it at 32 or even 16 is probably more sane than 64. (at least for a new project)  
It really just boils down to whether we care about keeping the old timing accessible even though it wasn't good.

somewhatlurker:
Hmm...  I realised I didn't make mixers run more often than the default (5ms?) when the HID period is decreased. That might be a good idea.  
Probably not strictly necessary, but I should've had the foresight to copy that stuff.

TheRealMoeder:
I took another T8SG lying around (actually a Radiomaster TX8, but they are all the same to me ;) and flashed it. Now I'm not so sure anymore about the Wireshark measurements. I always get 11.25 ms no matter what setting I have in protocol options. Not sure what I saw in my last measurements, admittedly they were made in a rush.

TheRealMoeder:
Finally managed a working modular build for the T8SG. Works very well with the same behavior as non modular builds. Let's decide on the period setting and we're good for merging.

TheRealMoeder:
Concerning the protocol option:
Using steps of 2 isn't really good, as it still doesn't do what we expect it to do. May be the hackish
`static const char * const usbhid_opts[] = {
  _tr_noop("Period (ms)"),  "32", "16", "2", "4", "8", NULL,
  NULL
};` gives the best user experience defaulting on 8 ms.

somewhatlurker:
I think it's best to keep everything in a sequential order here, such as `"16", "8", "4", "2", "1"`, or even changing it to Hz for ascending too (`"62.5", "125", "250", "500", "1000"`).

I think the first option will become the true default when all is complete, right? (I don't see any way to set an alternate default without breaking the first listed option unless we write a callback in the protocol)

TheRealMoeder:
The issue is that the last entry is the default...

somewhatlurker:
It doesn't seem to behave that way on other protocols in emulator.
For example, the default radio in testrf is CYRF6936 for me.

I suspect you accidentally loaded a model that already had config data or left in the code that sets it to 8 when it was previously unset.

TheRealMoeder:
Reading up on USB... Actually the period depends on the operating speed of the USB device and host. On full and high speed we can have 1,2,4,8,16,32 but for low speed it is only 8, 16 or 32. So setting period in protocol interval doesn't necessarily reflect the true transfer interval of the hid.

somewhatlurker:
I believe all targets with HID support are running at full speed, and it would be extremely unlikely for someone to use a host device without full speed support.

TheRealMoeder:
I don't see tooany compatibility issues as USB is well defined, so what do you think of just making 8 ms the slowest and also default value with option of turning up to 1 ms? Mixers are running slower anyways right now,if I'm not mistaken.

About the order of protocol options, it is just what I observed. I was surprised by the behavior and also tested with a new model, but got the same behavior. Maybe this is related to the code in lines 117/118? Do we really need them?

somewhatlurker:
Yep, that's what I meant by `the code that sets it to 8 when it was previously unset` -- those lines and 99 were used to set a default value when the option was set to 0. Now with text options, they just override the option at index 0 to index 8 and will need to be removed.

somewhatlurker:
And yeah, if this is going to be more trouble than it's worth to get more periods added in a way that makes sense, 8ms or less sounds fine.

I'll get an update ready later with options as 125,250,500,1000 Hz (making use of the defined options to display these in a unit users are probably more familiar with and a nice ascending order) and extra mixer calls as necessary using the same logic as sbus/sumd.

TheRealMoeder:
Sounds good. I'll merge as soon as it is finished and looks good. Can you please also add a line in deviation-manual to explain the protocol option ?

src/protocol/usbhid.c
```diff
@@ -57,13 +78,40 @@ static void build_data_pkt()
...
+    u16 protoopts_period = period_index_to_ms(Model.proto_opts[PROTO_OPTS_PERIOD]);
+    if (usbhid_period_ms != protoopts_period) {
+        usbhid_period_ms = protoopts_period;
+        // HID should be restarted when period changes
+        // this lets us update the endpoint descriptor's bInterval field
+        HID_Disable();
+        HID_SetInterval(usbhid_period_ms);
+        HID_Enable();
+    }
+
```
PhracturedBlue, line 100: Lint Error:
```
100:  Line ends in whitespace.  Consider deleting these extra spaces.  
```

somewhatlurker:
Okay, can you please test this on hardware again and I'll go look into the manual stuff.

TheRealMoeder:
Yes will do so, but not before tomorrow night. @Goebish can you also review the code again? Protocols are your domain 😉

TheRealMoeder:
You're lucky, I found a minute to try it out. Works fine on hardware, although I still once in a while get those odd 11.25 ms measurements on my MacOS machine. But I doubt this really is a deviation issue, rather something with Darwin or Wireshark. 1 ms was verified as well, test fly also successful. I'll give a green light, maybe someone on a windows machine can also try it out.

somewhatlurker:
Yeah, technically speaking to get perfect timing you'd need to generate packets on demand as the interrupt fires.
Looking at it now, it could be done by waiting for `usb_preXferComplete` to be 1 before trying to write anything.

Something like this:
```
    if (!usb_preXferComplete) return 100;  // wait until endpoint is ready
    switch (state) {
        case ST_DATA1:
            CLOCK_RunMixer();    // clears mixer_sync, which is then set when mixer update complete
            state = ST_DATA2;
            return mixer_runtime;

        case ST_DATA2:
            if (mixer_sync != MIX_DONE && mixer_runtime < 2000) mixer_runtime += 50;
            build_data_pkt();
            HID_Write(packet, sizeof(packet));
            state = ST_DATA1;
            return usbhid_period_ms * 1000 - mixer_runtime - 500;  // return with - 500 in case host is polling faster than our clock
    }
    return usbhid_period_ms * 1000 - 500;   // avoid compiler warning
```

Or alternatively, with changes to allow handling the USB endpoint callback directly, but I'm not sure we can run mixers fast enough to definitely not cause issues on the bus from that.

somewhatlurker:
let's see if this actually improves anything...
might actually reduce battery life a little if it works as expected, but it'll be fine.. USB HID should be using less power than most protocols by not running any RF anyway.

TheRealMoeder:
While this might make  a little improvement, I'm a little concerned in yet another rom size increase in the modular platforms (especially devo7e). These are really tight on rom space, and the addition of HID_setinterval already adds a bit. It might even be useful to make thr default option of 8 ms hard-coded for the modular builds and add the protocol option, which needs HID_SetInterval, only in non modular builds

somewhatlurker:
Hmm..  I see...
I really have no good sense for what's significant or not at these sizes, but I'll see if I can find some way to get it back down before removing functionality.

TheRealMoeder:
You last commit didn't add too much, it's just something we (unfortunately) always have to have in mind.

somewhatlurker:
Yeah, I get what you mean. It's easy to add things that use up space now, but if we always do that it becomes hard to get the space later for something that really needs it.

TheRealMoeder:
We put a lot of effort into code cleanup, because ROM space for devo7e was absolutely used. Since then, things have gotten a little relaxed (and deviation development more or less subsided due to not many active developers). But as you notice in target_defs, a lot of features are already deactivated for the modular builds.

somewhatlurker:
Looks like devo 8 is getting pretty close to fully used too, and t8sg not far behind...  I suppose the only way to avoid the increase on non-modular is to simply not add code though right?

somewhatlurker:
Anyway, the increase seems pretty minimal for 7e now, so if that was the primary concern and the code is still working, I guess we're good.

somewhatlurker:
Hmm...   I think I actually figured out a nice enough way to give us a callback without needing to run mixers in the actual USB endpoint callback -- the USB endpoint callback can start a clock to run the protocol callback in an interrupt later.
That should be simple to implement, and hopefully be reliable.

somewhatlurker:
I'll play around tomorrow and see if I can get it to compile.

somewhatlurker:
Hopefully I didn't just break everything lol

somewhatlurker:
Perhaps this PR could use a revised title due to the now larger scope of changes, and possibly incorporate an increase to 8 button inputs with it being more than a simple change now anyway.
Or should it stay limited to just timing-related changes and be followed up by a smaller PR for more channels later?

TheRealMoeder:
I'd prefer seperate Pull requests for timing and inputs.

TheRealMoeder:
99fb44b broke USBHID for modular as well as non-modular builds, while 5c0a6bd6 works for both.

somewhatlurker:
Does the callback method work with 1 instead of 0 in the CLOCK_StartTimer call on devo_hid.c line 146?

TheRealMoeder:
Nope.

somewhatlurker:
Hmm...  I'd like to see the callback method working, but I can't see what I'm doing wrong.
Should I just revert to what works for now?

TheRealMoeder:
Yes, I also expected it to work, but haven‘t spent time debugging. Since none of the the protocols currently runs mixers on demand and the benefit is small and should only noticeable with 500 Hz and above, I‘m fine with the previous solution.

somewhatlurker:
Running mixers on demand was already present since before 5c0a6bd. It was copied from sbus/sumd like a lot of other stuff. The callback from USB code was just to try guaranteeing perfect sync to USB host polling.

I will go ahead and revert the two most recent commits for the sake of having a working system though.

TheRealMoeder:
Yes I know, I didn't mean running mixers on demand but synchronizing callbacks. Mixer run time is about 1 ms worst case, I'm not too worried about this. #532 is some interesting read.

TheRealMoeder:
I can't approve as I created this PR, but I'm good with this.

goebish (approved):

Issue #999, opened 2020-09-14
[Bug?] Custom voice alerts assume contiguous and ascending ordering in voice.ini


somewhatlurker:
Not necessarily a huge issue, but I noticed it when playing around in emulator.  
Custom alerts in `voice_map` seem to be created by writing to indexes 200+ in the order alerts are found in the file, not corresponding to the key in voice.ini.  
While the firmware seems to handle this fine internally (the order of alerts in the selector changes, but they're played using the correct id from voice_map), model files just store the internally used index and will break if voice.ini is merely reordered.

For example, two voice.ini files containing `200=0,300\n201=1,300` and `201=1,300\n200=0,300` in the custom section are handled differently. (In the second case, alert 200 in model files is actually `1` (mp3 file 201) and alert 201 is `0` (mp3 file 200))  
It's a very non-intuitive behaviour and I assume it's not intentional.

　

It seems pretty simple to fix; ids in the model file should just have to be converted to the proper internal id:

when saving: (src/config/model.c line 1385 -- near `fprintf(fh, "[%s]\n", SECTION_VOICE);` ?)
look up the correct id of music in voice_map

when loading: (src/config/model.c line 1030 -- near `if (MATCH_SECTION(SECTION_VOICE)) {` ?)
iterate on voice_map until an entry with id == val is found (a little slow, but probably fine)
if no match return 0

eried:
That is why I just append new voices to the end and have a excel to create the file just in case. But it would be nice if the file supported non-contiguous

TheRealMoeder:
I was never overly satisfied with the whole voice handling, but it grew out of limits due to unidirectional communication with dfplayer and ram limitations on some targets. If we use internal id's in the model files, those can't be edited manually anymore, which is also not a good solution.

somewhatlurker:
The current implementation already uses internal ids in model files, they just happen to look like the real mp3 ids. Manual editing shouldn't get any harder after the change.

Although on the topic of causing new issues, I realise now that that naive fix could screw up model files for users who have non-standard edited voice files that unknowingly hit this bug.  
That could be fixed by being able to read two types of value, such as `HOLD0=200` for old loading behaviour and `HOLD0=mp3-200` for new loading behaviour -- then old config can't be broken (I'm not sure what an appropriate prefix would be though; `mp3-`  or `ini-` seem appropriate, but aren't very descriptive)

somewhatlurker:
[...]

PR #1000, opened 2020-09-16
Save proper external voice IDs in model files

somewhatlurker:
This is one possible fix for #999.

If merged, model files will now save the alert IDs from voice.ini/mp3 file names for custom alerts, instead of the internal index in voice_map.

To ensure old models are loaded correctly, an "mp3-" prefix is added to alert values saved in the new way (all models will be saved with new IDs).  
This does have a side effect of making new model files incompatible with old firmware though. Alternatives would be to either not use a prefix and make it possible for old models to load with incorrect IDs when used with a poorly formatted voice.ini (probably not desirable) or use a prefix/suffix that atoi won't error out on (eg, "+" prefix), although that would make the file less readable without writing an explanation comment.

### DIFF
--- a/src/config/model.c
+++ b/src/config/model.c
@@ -243,7 +243,7 @@ static const char * const VOICE_TELEMALARM[TELEM_NUM_ALARMS] =
     { "telemalarm1", "telemalarm2", "telemalarm3", "telemalarm4", "telemalarm5", "telemalarm6" };
 static const char * const VOICE_TIMER[NUM_TIMERS] =
     { "timer1", "timer2", "timer3", "timer4" };
-
+static const char VOICE_MP3ID_PREFIX[] = "mp3-";
 #endif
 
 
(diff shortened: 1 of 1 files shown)


src/config/model.c
```diff
@@ -1075,15 +1075,32 @@ int assign_int(void* ptr, const struct struct_map *map, int map_size)
...
+                if (voice_map[i].id == val) {
+                    idx = i;
+                    break;
+                }
+            }
+        }
+        else {
+            idx = val;
+        }
+        if(idx > MAX_VOICEMAP_ENTRIES-1 || voice_map[idx].duration == 0 || idx < CUSTOM_ALARM_ID) {
```
PhracturedBlue, line 1096: Lint Error:
```
1093:  If an else has a brace on one side, it should have it on both  
1096:  Missing space before ( in if(  
```

src/config/model.c
```diff
@@ -1433,29 +1450,29 @@ u8 CONFIG_WriteModel(u8 model_num) {
     fprintf(fh, "[%s]\n", SECTION_VOICE);
     for (idx = 0; idx < NUM_SWITCHES; idx++) {
         if (m->voice.switches[idx].music)
-            fprintf(fh, "%s=%d\n", INPUT_SourceName(file,idx + INP_HAS_CALIBRATION + 1), m->voice.switches[idx].music);
+            fprintf(fh, "%s=%s%d\n", INPUT_SourceName(file,idx + INP_HAS_CALIBRATION + 1), VOICE_MP3ID_PREFIX, voice_map[m->voice.switches[idx].music].id);
```
PhracturedBlue, line 1453: Lint Error:
```
1453:  Missing space after ,  
```

src/config/model.c
```diff
@@ -1433,29 +1450,29 @@ u8 CONFIG_WriteModel(u8 model_num) {
...
     }
 #if NUM_AUX_KNOBS
     for (idx = 0; idx < NUM_AUX_KNOBS * 2; idx++) {
         if (m->voice.aux[idx].music) {
             if (idx % 2)
-                fprintf(fh, "%s_UP=%d\n", INPUT_SourceName(tempstring,(idx-1) / 2 + NUM_STICKS + 1), m->voice.aux[idx].music);
+                fprintf(fh, "%s_UP=%s%d\n", INPUT_SourceName(tempstring,(idx-1) / 2 + NUM_STICKS + 1), VOICE_MP3ID_PREFIX, voice_map[m->voice.aux[idx].music].id);
             else
-                fprintf(fh, "%s_DOWN=%d\n", INPUT_SourceName(tempstring,idx / 2 + NUM_STICKS + 1), m->voice.aux[idx].music);
+                fprintf(fh, "%s_DOWN=%s%d\n", INPUT_SourceName(tempstring,idx / 2 + NUM_STICKS + 1), VOICE_MP3ID_PREFIX, voice_map[m->voice.aux[idx].music].id);
```
PhracturedBlue, line 1461: Lint Error:
```
1459:  Missing space after ,  
1461:  Missing space after ,  
```

TheRealMoeder:
From a quick glance this looks good. I actually like the idea of using the mp3-prefix. This could also be the way to enhance this to add a "beep" prefix to set certain alerts to the beeps defined in sound.ini and making that somehow configurable in the voice config menu.

PR #1013, opened 2021-08-07
fix crsf datatype for config page

StonedDawg:
Just a very small fix of the CRSF config page when showing int16 and uint16 data types.


goebish (approved):

howels:
Please merge when convenient - this will help users of the popular Express LRS protocol as well as Crossfire/Tracer users.

PR #1015, opened 2021-09-26
Add XM+ bind fix only for non modular targets

MJ666:


howels:
Note: This should fix nightly builds on flash-constrained targets.

howels:
Please can someone review this to get nightly builds rolling again?

goebish (approved):

howels:
Thanks @goebish

PR #1020, opened 2022-01-09
Fix incorrect handling of long names

hexfet:
Names of max length (16 including nul) or greater caused incorrect calculation of location of next parameter in receiver buffer.

Fixes issue in serial sending code where link was switched from tx to rx before last byte finished sending.

Fixes updating CRSF config page device list when new info message received from new device.

### DIFF
--- a/src/crsf.h
+++ b/src/crsf.h
@@ -22,10 +22,14 @@
 
 // Frame Type
 #define TYPE_GPS              0x02
+#define TYPE_VARIO            0x07
 #define TYPE_BATTERY          0x08
-#define TYPE_VIDEO            0x08
+#define TYPE_HEARTBEAT        0x0b
+#define TYPE_VIDEO            0x0F
 #define TYPE_LINK             0x14
 #define TYPE_CHANNELS         0x16
+#define TYPE_RX_ID            0x1C
+#define TYPE_TX_ID            0x1D
 #define TYPE_ATTITUDE         0x1E
 #define TYPE_FLIGHT_MODE      0x21
 #define TYPE_PING_DEVICES     0x28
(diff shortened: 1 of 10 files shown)


howels:
Unfortunately this makes no difference. Device names longer than 14 characters are still shown as blank lines, same as 2b63b49 - your last PR for mixer-sync / CRSFshot

Without RX connected:


With RX:


With RX and shorter name

hexfet:
Thanks for the testing.  Since I don't have an ELRS module I simulate the expected message from the module with the longer name, and it does display and work correctly.  That suggests something is different about the message coming from ELRS.  Do you have the ability to capture the serial data coming from the ELRS module to Deviation?  That would be very helpful.   I'll also take a look at the ELRS code to see if I can spot anything.

howels:
Can you post your T8SGv2-plus build please? I am building 75cd90e09ac2d0f4908c168cacdad1624c08834a but having no luck for some reason.

howels:
@StonedDawg is there any way to dump the menu data via serial debug from the eLRS TX?

hexfet:
Here's a build that also includes the crsfshot-sync change.
[deviation-t8sg_v2_plus-v5.0.0-75cd90e.zip](https://github.com/DeviationTX/deviation/files/7857129/deviation-t8sg_v2_plus-v5.0.0-75cd90e.zip)

howels:
I don't know what to say. I have flashed https://github.com/DeviationTX/deviation/commit/d0ca7634e06c9877f31c20ed4f126edd2e87770e which you posted to the DeviationTX forums. I am building eLRS from source and uploading via the ExpressLRS-configurator tool. If I use the source as provided, I see no TX device. If I edit this line https://github.com/ExpressLRS/ExpressLRS/blob/32b5077d00be869978444b5aa30a6952ac9d3bc2/src/include/target/BETAFPV_2400_TX_MICRO.h#L2 and shorten the name by a few characters, then the TX appears.
I can see the string doca763 on boot so I know it flashed ok. I am seeing if I can get serial output but simply enabling debug won't help as that only outputs the radio link stats and not menu data.

mooiweertje:
Is everybody using ExpressLRS-configurator 1.3.3?

howels:
As pointed out by @mooiweertje this only works at <500Hz. On 500Hz the original problem persists.

howels:
At 500Hz there is a small pause between bytes as stick commands are sent during the device-info transfer (cannot be chunked). OpenTX handles this pause OK, but sounds like on DeviationTX this pause is causing the transfer to fail and the data currently in the buffer is not returned.

hexfet (commented):
Tested by users mooiweertje and belrik on deviationtx.com forum.

PR #1019, opened 2022-01-08
Implement CRSFshot synchronization between tx and radio

hexfet:
Use CRSFshot "offset" telemetry to synchronize packet sends with the
transmitter module.  This reduces latency by delivering data to the
transmitter in sync with when the transmitter will use it.

Prior to this change a Crossfire TX LCD shows "CRSF V2" for the RC
Input.  After this change the display is "CRSFshot".

PR #1018, opened 2021-11-13
Added crsf-shot

phobos-:
Implementation is based on opentx, tested with ExpressLRS hardware.
Before - packet rate is stuck at 250 packets per second.
After - packet rate depends on the settings and varies from 25 to 500 per second.

Tested with T8SG V2 plus, builds for T8SG V2 and T8SG V2 plus attached below.
[deviation-t8sg_v2_plus-v5.0.0-f04bc27.zip](https://github.com/DeviationTX/deviation/files/7532297/deviation-t8sg_v2_plus-v5.0.0-f04bc27.zip)
[deviation-t8sg_v2-v5.0.0-f04bc27.zip](https://github.com/DeviationTX/deviation/files/7532298/deviation-t8sg_v2-v5.0.0-f04bc27.zip)

PR #1023, opened 2022-01-21
CRSF updates for ELRS

hexfet:
* Add protocol ELRS. Same source code as CRSF but with serial datarate 1.87Mbps
* Add CRSF model match message. Set fixed id 0-63 to use model match.
* Add support for ELRS info message.  Popup dialog when Model Mismatch reported.  The ELRS info message is only requested when disarmed so as not to interfere with stick packets during flight.
* Workaround for glitch when scrolling non-selectable items
* Serial send and receive refactored with additional buffer layer to handle higher data rate


phobos-:
This is awesome work @hexfet . I am testing it and so far so good! 👍

hexfet:
Thanks!  Appreciate the testing @phobos- .  Just made a PR to ELRS that helps with the UART errors if you could try it.

howels:
@hexfet is this working with the OPENTX_SYNC feature in expressLRS?

hexfet:
@howels Yes, PRs #1018 and #1019 brought in support for the sync packets.

hexfet:
Tested by deviationtx.com user mooiweertje and myself.

PR #1024, opened 2022-02-19
Add configurable serial bitrate to CRSF protocol

hexfet:
When clicking on the CRSF protocol name the current model name will appear at the bottom of the device list. Clicking on the model name allows the bit rate to be set. I chose to display the model name to emphasize the bit rate is stored separately for each model.

The available bitrates are 400K, 1.87M, and 2.25M. The ELRS PR #1435 is needed to run 2.25M, or just add 2.25M to the baudrates array in the ELRS code. This is the fastest I could get the serial link to run.

The ELRS protocol is no longer needed and is removed.  Model files saved with ELRS will be loaded with None as protocol.  Use CRSF and set the bitrate to 1.87M or 2.25M depending on what ELRS supports.


hexfet:
Several users have reported this change works well.

PR #1050, opened 2025-03-15
Support alternate arming method in ELRS version 4

hexfet:
ELRS version 4 will be adding a new arming method that doesn't use channel 5 (CH5) for arming.  When the alternate method is used channel 5 can be used for full range in full-res modes.  Details are available [here](https://github.com/ExpressLRS/ExpressLRS/pull/3008).

This change adds a CRSF protocol option to support the new arming method.  In the CRSF config model device the new option is "ELRS Arm".  The default value is CH5 to maintain backward compatibility - CH5 is used for arming the same as in previous ELRS versions.  The option can be changed to select the arming channel as any virtual channel (Virt1-Virt10).  Then the virtual channel mixer output becomes the arming channel.  Set the virtual channel mixer to output greater than zero to arm, less than zero to disarm.  For example use the default simple mixer and set the Src to the switch to be used for arming.  Note that changing the mixer name (e.g. "Virt1" to "ARM") will not change the channel name displayed in the protocol options.

The protocol option only has an effect if the radio module is running ELRS version 4.0.0 or later.  The option is ignored for ELRS v3 where CH5 is the only arming method supported.


TheRealMoeder (approved):

TheRealMoeder (approved):

TheRealMoeder (approved):

hexfet:
@TheRealMoeder thanks for the reviews and approvals!

Tested with ELRS build master@{225f943b}

PR #1052, opened 2025-06-26
Fix crash when opening telemetry monitor for FrSky

gBaltar:
Telemetry monitor did only show a single box centered and was crashing my F12e seconds later.


hexfet:
Thanks for the contribution!

